PR #6, opened 2018-09-09
Add vpc flow logs option to subnets

andreyk-code:
Added enable VPC flow logs option to module and examples


main.tf
```diff
@@ -34,6 +34,7 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range            = "${lookup(var.subnets[count.index], "subnet_ip")}"
   region                   = "${lookup(var.subnets[count.index], "subnet_region")}"
   private_ip_google_access = "${lookup(var.subnets[count.index], "subnet_private_access")}"
+  enable_flow_logs         = "${lookup(var.subnets[count.index], "subnet_flow_logs")}"
```
morgante: @andreyk-code We should add a default (`false`) here so it's not required to pass. Also should document this in the README.

andreyk-code: Good point! Added in latest commit and updated README. Also made the same change for "private_ip_google_access" value.

PR #8, opened 2018-09-13
Add the created network URI as an output

asproul:
This change adds the `self_link` of the created network as an output. Also updated the examples and generated the READMEs with terraform-docs.


morgante:
LGTM, thanks!

PR #11, opened 2018-09-18
Updated terraform-provider-google min version to 1.12.0

rbramwell:
Overview:
- The enable_flow_logs field to subnetwork resource was introduced in terraform-provider-google version 1.12.0. Used at https://github.com/rbramwell/terraform-google-network/blob/master/main.tf#L37
- See https://github.com/terraform-providers/terraform-provider-google/blob/master/CHANGELOG.md#1120-may-04-2018


morgante:
Thanks for the contribution.

PR #13, opened 2018-09-24
Update subnet_flow_logs and subnet_private_access to strings from boolean

andreyk-code:


examples/multi_vpc/main.tf
```diff
@@ -98,4 +124,30 @@ module "test-vpc-module-02" {
 
     test-network-02-subnet-02 = []
   }
+
+  fw_rules_ingress_with_tags = [
```
morgante: @andreyk-code Do these have to be separated out like this?

andreyk-code: I don't think so. Trying to control the firewall type within the actual resource is difficult with current interpolation syntax. Since the firewall rule resource uses different inputs depending on if it's an ingress/egress and if it uses as network tags or service_accounts. It can't have both source ranges/destination ranges or accounts/tags, and sending a null resource isn't possible. This removes ability for such logic:

```
  fw_rules = [
    {
      rule_name       = "allow-icmp"
      type = "service_account"
      direction = "egress"
      ...
    }
]
```
Doing it through count is challenging since we can't nullify the resource based off previous logic. Something like this would error since we can't track count outside of the resource:

```
  count = "${lookup(var.fw_rules_ingress_with_tags[count.index], "type", "") == "tags,ingress" ? "${var.fw_rules_ingress_with_tags}" : 0 }"
```

Maybe it's not worth trying to add firewall rules until v0.12 is released? It's fairly convoluted and almost simpler to just create independent firewall resources in these versions. In 0.12, it will become way cleaner since we can null out inputs:

```
  fw_rules_allow = [
    {
      rule_name     = "example-rule-01"
      direction     = "INGRESS"
      target_tags   = "example test"
      source_ranges = "0.0.0.0/0"

      icmp = true
      tcp  = true

      tcp_ports = "22 80"
    },
  ]
```
```
resource "google_compute_firewall" "allow-rule" {
  count = "${length(var.fw_rules_allow)}"
  ...
  direction               = "${lookup(var.fw_rules_allow[count.index], "direction")}"
  source_ranges           = "${lookup(var.fw_rules_allow[count.index], "direction") == "INGRESS" ? split(" ", lookup(var.fw_rules_allow[count.index], "source_ranges", " ")) : null}"
  destination_ranges      = "${lookup(var.fw_rules_allow[count.index], "direction") == "EGRESS" ? split(" ", lookup(var.fw_rules_allow[count.index], "destination_ranges", " ")) : null}"
```

andreyk-code: Done, pulled out firewall support

andreyk-code:
Updated PR title from "Add firewall support and update subnet_flow_logs, subnet_private_access to strings from boolean" to "Update subnet_flow_logs and subnet_private_access to strings from boolean"

PR #16, opened 2018-10-19
Support no secondary ranges to be defined

efouarge:
Update to the module to support no secondary ranges to be defined

- [x] if it's possible to have secondary_ip_range default to an empty list in cases where it's not needed.
- [x] `secondary_ranges = {}` set as a default value, and confirm that the doc generator remains
- [ ]  Restructure examples to simple, multi-vpc, and complex to support the new structures


examples/simple_project_no_secondary/main.tf
```diff
@@ -0,0 +1,38 @@
...
+    {
+      subnet_name           = "subnet-02"
+      subnet_ip             = "10.10.20.0/24"
+      subnet_region         = "us-west1"
+      subnet_private_access = "true"
+      subnet_flow_logs      = "true"
+    },
+  ]
+
+  secondary_ranges = {}
```
morgante: Does this empty map need to be provided? Can't that be the default value for the variable?

examples/simple_project_no_secondary/main.tf
```diff
@@ -0,0 +1,38 @@
+/**
```
morgante: Probably we could compress this into the simple project example and move secondary ranges into a "complex" example.

main.tf
```diff
@@ -41,6 +41,18 @@ resource "google_compute_subnetwork" "subnetwork" {
   secondary_ip_range = "${var.secondary_ranges[lookup(var.subnets[count.index], "subnet_name")]}"
 }
 
+resource "google_compute_subnetwork" "no_secondary_subnetwork" {
```
morgante: Is it possible to collapse these into a single resource? Having separate resources has major implications for state as it means adding a secondary range to a subnet would be impossible (it would attempt to delete the subnet, then create it fresh).

examples/simple_project_no_secondary/main.tf
efouarge: I will fold that into it and restructure some of the examples accordingly

examples/simple_project_no_secondary/main.tf
efouarge: We could move it to a default variable, I need to play with this to see what impact this will have with the doc generator

main.tf
morgante: > Adding secondary networks post creation from my dealings has not been a common occurrence and is only used within private GKE clusters in shared-vpc. 

It's not especially uncommon though, as you will occasionally need to start using GKE on an existing network.

> Currently the way the module is constructed it's forcing to burn quite a bit of your supernet scopes. So the likely transition for this would be to create the new segment in the network and drain your resources over to the new networks.

That's still pretty suboptimal. Often enterprises have constrained subnet space so needing to create a whole new primary subnet to introduce secondary ranges isn't ideal. Could you do a bit more discovery here? I'm wondering if it's possible to have `secondary_ip_range` default to an empty list in cases where it's not needed.

efouarge: > Could you do a bit more discovery here? I'm wondering if it's possible to have secondary_ip_range default to an empty list in cases where it's not needed.

Yeah I can look into that further

aaron-lane:
@efouarge are you available to address the review?

morgante:
@skenmy Could you take a look at this and possibly pick up the work? With the 0.12 refactor, I think the approach might have changed quite a bit.

skenmy:
ack - assigned myself

skenmy:
@morgante can you take a look at what I did that broke builds / tests on GitHub?

morgante:
@skenmy Tests should be working now, just push a new commit to test.

bmenasha:
also see my prior pull request for what i think is a bit better syntax for passing this list through:
thanks

https://github.com/terraform-google-modules/terraform-google-network/pull/62/files#diff-7a370d8342e7203b805911c92454f0f4R50

skenmy:
@morgante looks like still some issues with the builds? LMK if I need to do anything!

morgante:
@skenmy Can you push a commit to trigger the run again? I think permissions were off.

main.tf
```diff
@@ -47,7 +47,7 @@ resource "google_compute_subnetwork" "subnetwork" {
   enable_flow_logs         = lookup(var.subnets[count.index], "subnet_flow_logs", "false")
   network                  = google_compute_network.network.name
   project                  = var.project_id
-  secondary_ip_range       = var.secondary_ranges[lookup(var.subnets[count.index], "subnet_name", null)]
+  secondary_ip_range       = [for i in range(length(contains(keys(var.secondary_ranges), var.subnets[count.index]["subnet_name"]) == true ? var.secondary_ranges[var.subnets[count.index]["subnet_name"]] : [])) : var.secondary_ranges[var.subnets[count.index]["subnet_name"]][i]]
```
tfhartmann: I'm not sure this is the right PR for this comment, but  I wanted to raise an idea.  Please feel free to let me know if I should raise this in an issue instead 😄    Rather then use a for loop here, what about the idea of using a dynamic block, and getting rid of the `secondary_ranges` data structure all together, and replacing it with a `secondary_ranges` list of maps in the `subnets` map?

morgante: I think that's a good idea and would be happy to look at a PR.

PR #19, opened 2018-11-29
Adds subnet data source, allows no secondary range in subnets; fixes #5

### LINKED ISSUE #5: Creating a network with no secondary ranges fails
The output expects at least one subnet to have a secondary range.

Config:
```hcl
module "network2" {
  source       = "../../modules/network"
  project_id   = "${data.terraform_remote_state.projects.net-host-project-id}"
  network_name = "sandbox-vpc2"
  shared_vpc_host = "true"

  subnets = [
    {
      subnet_name           = "sandbox-compute"
      subnet_ip             = "10.18.0.0/20"
      subnet_region         = "us-central1"
      subnet_private_access = false
    },
  ]

  secondary_ranges = {
    sandbox-compute = []
  }
}
```

Error:
```
Error: Error running plan: 1 error(s) occurred:

* module.network2.output.subnets_secondary_ranges: Resource 'google_compute_subnetwork.subnetwork' does not have attribute 'secondary_ip_range' for variable 'google_compute_subnetwork.subnetwork.*.secondary_ip_range'
```

angstwad:
This will fix #5, which causes errors when the `secondary_ranges` map does not define any secondary ranges.  This keeps the `subnets_secondary_ranges` output, but we might consider removing it altogether, since it is capable of producing such useless output as:

```
Outputs:

secondary_ranges = [
    [
    ],
    [
    ],
    [
    ]
]
```

Removing the output is something I'm happy to do, if that's the direction we choose to go.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -49,6 +49,14 @@ resource "google_compute_subnetwork" "subnetwork" {
   secondary_ip_range = "${var.secondary_ranges[lookup(var.subnets[count.index], "subnet_name")]}"
 }
 
+data "google_compute_subnetwork" "created_subnets" {
+  count = "${length(var.subnets)}"
+
+  name    = "${lookup(var.subnets[count.index], "subnet_name")}"
+  region  = "${lookup(var.subnets[count.index], "subnet_region")}"
+  project = "${var.project_id}"
+}
+
 /******************************************
 	Routes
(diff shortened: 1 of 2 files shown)


morgante:
Thanks for the contribution!

PR #21, opened 2019-01-10
Remove secondary networks from simple example

morgante:


main.tf
```diff
@@ -49,6 +49,14 @@ resource "google_compute_subnetwork" "subnetwork" {
   secondary_ip_range = "${var.secondary_ranges[lookup(var.subnets[count.index], "subnet_name")]}"
 }
 
+data "google_compute_subnetwork" "created_subnets" {
+  count = "${length(var.subnets)}"
+
+  name    = "${lookup(var.subnets[count.index], "subnet_name")}"
```
aaron-lane: Should this use the `google_compute_subnetwork` resource rather than the variables to ensure the subnetworks are created before being accessed?

PR #26, opened 2019-01-28
Add ability to choose network routing mode

walterdolce:
This PR adds the ability to specify the network routing mode, defaulting to `GLOBAL` (as it is now).


andreyk-code:
Thank you for contributing @walterdolce! 

Changes look good. Before we merge, could you please add the other value that is possible, "regional", to the input description in readme. Alternatively, could add an example to examples/ with regional VPC setting.

andreyk-code:
Thank you! LGTM

walterdolce:
Thank YOU!

PR #27, opened 2019-01-29
Add subnet_self_links as outputs

pratikmallya:
We ran into this when using the `google_compute_router_nat`
tf module. This requires the subnet self links as inputs when
defining the NAT on a per subnet basis


pratikmallya:
@morgante if possible, would appreciate a quick review

PR #33, opened 2019-03-01
Add fully-qualified subnet IDs as output

NFollett89:
The Project Factory module requires the fully-qualified subnet IDs to be passed in when using a Shared VPC.

The best user experience involves pulling a data source from the Network module's remote state to fetch this value dynamically, and until this change was not offered as an output.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -39,6 +39,11 @@ output "subnets_self_links" {
   description = "The self-links of subnets being created"
 }
 
+output "subnets_ids" {
+  value       = "${formatlist("projects/%s/regions/%s/subnetworks/%s", var.project_id, google_compute_subnetwork.subnetwork.*.region, google_compute_subnetwork.subnetwork.*.name)}"
+  description = "The fully-qualified IDs of subnets being created. Accepted by some Google modules"
+}
+
 output "subnets_regions" {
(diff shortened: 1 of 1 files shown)


morgante:
I'm not sure about this, seems like the self-links should already contain all the required information?

NFollett89:
@morgante The self-links do contain all the information, but they are prefixed with a bunch of text that is not compatible with what the Project Factory requires when adding subnet membership.

Self link: `https://www.googleapis.com/compute/v1/projects/PROJECT/regions/REGION/subnetworks/SUBNET`

FQ Subnet ID: `projects/PROJECT/regions/REGION/subnetworks/SUBNET`

I tried using the self links at first and it produces an error.
```
googleapi: Error 400: Invalid value for field 'region': 'compute'. Unknown region., invalid
```

The failure happens [here](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L222), among other places.  The strings are tokenized into different list lengths, and so using an index of 3 and 5 to access certain values is breaking for the self link when it's expecting the other format.

morgante:
The right way to solve this is probably to update the Project Factory to accept self-links.

As an interim, you can strip the self-link.

PR #34, opened 2019-03-12
Refactor integration tests to include random names

glarizza:
This commit updates the integration test fixtures to allow for
randomized subnetwork names as well as randomized network names. Without
this commit the integration test fixtures will create VPC networks with
randomized names, but the subnetwork names will remain static. Because
subnetworks are regional resources, you can't have two subnetworks with
the same name tied to the same project and region. This means that
without adding randomness to subnetwork names, only 1 run of the
integration test pipeline can occur at any given time.

Normally a `random_string` resource is used to impart randomness into
resource names, but because of the way lookups are performed on the
`subnets` and `secondary_ranges` input variables, those variables cannot
contain computed values. To remedy this the test fixtures in
`test/fixtures/<scenario name>` now include an input variable named
`random_string_for_testing` that accepts a string to be appended on both
the network and subnetwork names. The `test/ci_integration.sh` script
has been modified to look for an environment variable named
`RANDOM_STRING_FOR_TESTING` and pass that to
`var.random_string_for_testing`. If that environment variable has not
been declared then a 5-character string is generated on the fly and
used.

This commit also contains an update to `Makefile` to ensure that
`test/ci_integration.sh` is sourced and the `setup_environment()`
function is called for any of the `docker_*` make targets. The
`Makefile` has also been modified to ensure that the environment
variables that are referenced by `test/ci_integration.sh` are being
passed down to the Docker container.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,50 +84,60 @@ docker_run:
 	docker run --rm -it \
 		-e CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${CREDENTIALS_PATH} \
 		-e GOOGLE_APPLICATION_CREDENTIALS=${CREDENTIALS_PATH} \
-		-e TF_VAR_project_id \
+		-e PROJECT_ID \
+		-e SERVICE_ACCOUNT_JSON \
+		-e RANDOM_STRING_FOR_TESTING \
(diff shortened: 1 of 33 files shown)


glarizza:
@aaron-lane Updated the strings into local values, pushed, and the tests are passing

PR #40, opened 2019-05-23
Initial import of net-firewall submodule from internal version.

ludoo:
This brings in the net-firewall module as a submodule.

I added a new example and tests, as I was not too sure that adding the submodule to the simple example was fine.

I'm not too sure about the version bump in `inspec.yml`, but it's the only way I could run verify witout Kitchen complaining about a version mismatch.

This PR also contains a minimal fix to the combine docfiles script, so it doesn't fail when a match is not found (as happens in this repo).

### DIFF
--- a/helpers/combine_docfiles.py
+++ b/helpers/combine_docfiles.py
@@ -54,10 +54,6 @@ groups = re.match(
 replace_content = groups[0] + groups[1]
 
 # Find where to put the replacement content, overwrite the input file
-m = re.match(insert_separator_regex, input, re.DOTALL)
-if not m:
-  print("insert regexp not matched in %s" % sys.argv[1])
-  sys.exit(0)
-groups = m.groups(0)
+groups = re.match(insert_separator_regex, input, re.DOTALL).groups(0)
 output = groups[0] + replace_content + groups[2] + "\n"
 open(sys.argv[1], "w").write(output)
--- a/helpers/combine_docfiles.py
+++ b/helpers/combine_docfiles.py
@@ -54,6 +54,10 @@ groups = re.match(
 replace_content = groups[0] + groups[1]
 
 # Find where to put the replacement content, overwrite the input file
-groups = re.match(insert_separator_regex, input, re.DOTALL).groups(0)
+m = re.match(insert_separator_regex, input, re.DOTALL)
+if not m:
+  print("insert regexp not matched in %s" % sys.argv[1])
+  sys.exit(0)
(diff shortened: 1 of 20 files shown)


ludoo:
Gentle ping, with a reminder that this code comes from an internal modules, has been used for months, and this is just a move to this repo as parte of the Fabric plan we drew together.

helpers/combine_docfiles.py
```diff
@@ -54,6 +54,10 @@
 replace_content = groups[0] + groups[1]
 
 # Find where to put the replacement content, overwrite the input file
-groups = re.match(insert_separator_regex, input, re.DOTALL).groups(0)
+m = re.match(insert_separator_regex, input, re.DOTALL)
```
aaron-lane: The bug related to this change has been recorded here https://github.com/terraform-google-modules/terraform-google-module-template/issues/16. The workaround is to ensure that the license comment is present in all Terraform configuration files.

test/integration/simple_project/inspec.yml
```diff
@@ -2,7 +2,7 @@ name: simple_project
 depends:
   - name: inspec-gcp
     git: https://github.com/inspec/inspec-gcp.git
-    version: ~> 0.9.0
+    version: ~> 0.11.0
```
aaron-lane: It turns out that the git profile fetcher does not actually support the `version` property: https://github.com/inspec/inspec/issues/2986.

The work around is to use `tag`.

```suggestion
    tag: v0.11.0
```

helpers/combine_docfiles.py
ludoo: Heh, the missing file is one unrelated to this CL, that's why I changed the script instead. I'll change the unrelated file and revert the script, thanks.

test/integration/simple_project/inspec.yml
ludoo: Hmmm ok, the change made the tests work, that's all I know. I'll try with tag.

helpers/combine_docfiles.py
ludoo: I fail to see how this is relevant, as the exception is raised in `./codelabs/simple/README.md`.

morgante: @ludoo It's fine if you leave as-is.

ludoo: Yup, I reverted the change, checks are fine. Feel free to approve. ;)

PR #42, opened 2019-06-02
Add support for shared vpc service project attachment

ludoo:
This adds support for attaching service projects to shared VPC, at the network or subnetwork level.

We need this for Fabric environments, where the logical place to declare service projects is at the network resource stage, instead of the project stage like project factory does (and we're not using project factory in fabric environments, but our simplified project submodule).

My strong preference would be to implement this into a separate submodule, which is unfortunately not possible due to the dependency on the `google_compute_shared_vpc_host_project` resource, and I could not find a way of doing that through outputs, as the resource only exposes the attributes that are passed in to it. If you have a sane way of doing it, let's explore it.

The other complexity derives from the fact that most of the variables needed for project atachment are dynamic when, as in Fabric environments, projects are created in the same root module: project ids, service accounts which are derived from project numbers, etc. This means that none of those variables can be used in counts, to avoid the dreaded "value of count cannot be computed" error, making variables design sub-optimal.

That said the code is not elegant but very simple (two resources with count), and is only used when `shared_vpc_host` is true. I have no way of testing it as it requires at least one more project, but I tested it locally and it works.

I'm open to other suggestions, but given the above constraints this looks like the best place to implement it. And of course, not having this available is blocking progress on the Fabric environments.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -35,16 +35,18 @@ resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
 
 resource "google_compute_shared_vpc_service_project" "projects" {
   count           = "${var.shared_vpc_host == "true" ? var.shared_vpc_service_projects_num : 0}"
-  host_project    = "host-project-id"
-  service_project = "${element(var.shared_vpc_service_projects, count.index)}"
+  host_project    = "${var.project_id}"
+  service_project = "${var.shared_vpc_service_projects[count.index]}"
   depends_on      = ["google_compute_shared_vpc_host_project.shared_vpc_host"]
 }
 
 resource "google_compute_subnetwork_iam_binding" "subnets" {
-  count      = "${var.shared_vpc_host == "true" ? length(var.shared_vpc_iam_subnets) : 0}"
-  subnetwork = "${element(var.shared_vpc_iam_subnets, count.index)}"
+  count      = "${var.shared_vpc_host == "true" ? length(var.shared_vpc_iam_subnet_names) : 0}"
+  project    = "${var.project_id}"
+  region     = "${lookup(var.shared_vpc_iam_subnets[count.index], "region")}"
+  subnetwork = "${var.shared_vpc_iam_subnet_names[count.index]}"
   role       = "roles/compute.networkUser"
-  members    = ["${split(",", element(var.shared_vpc_iam_members, count.index))}"]
+  members    = ["${split(",", lookup(var.shared_vpc_iam_subnets[count.index], "members"))}"]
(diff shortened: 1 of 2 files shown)


ludoo:
Actually there's a simpler way to trigger dependency, so that a submodule can be used to configure svpc attachments: just duplicate the `google_compute_shared_vpc_host_project` resource in the submodule. The `shared_vpc_host` in the main module can be true or false, and it won't matter as shared vpc will be enabled in the submodule. Let me know if this works better for you.

morgante:
@ludoo I would also prefer to do this via a submodule.

Could this work?
- Add an output to the root module:

```
output "shared_vpc_host_project_id" {
   value = "${google_compute_shared_vpc_host_project.shared_vpc_host.project_id}"
}
```

- Use an input in the submodule:

resource "google_compute_shared_vpc_service_project" "projects" {
  count           = "${var.shared_vpc_host == "true" ? var.shared_vpc_service_projects_num : 0}"
  host_project    = "${var.host_project}"
  service_project = "${var.shared_vpc_service_projects[count.index]}"
  depends_on      = ["google_compute_shared_vpc_host_project.shared_vpc_host"]
}

ludoo:
Heh, I tried that, but the project attribute is not computed, it looks like it's the same one that goes in so it's not creating an actual dependency.

I also much prefer doing this in a submodule, and actually don't mind the double resource, so if that's ok with you I'll close this and refactor as a submodule. Thanks!!!

morgante:
> Heh, I tried that, but the project attribute is not computed, it looks like it's the same one that goes in so it's not creating an actual dependency.

Really? Can you push up that code? Referencing *any* attribute from a resource should create a dependency, even if that attribute isn't computed.

My concern with the double resource is if you set it on both, destroy will get complicated.

ludoo:
It does create a dependency, but in my experience if the dependency is on an attribute which is not computed, it won't do what you expect.

morgante:
I don't follow. The host project resource would still be created first (if there's a dependency), no?

ludoo:
Hm, intuitively yes, but I had the same issue using project id on the project resource. Let me rework this as we will have a submodule, and do a few tests. Then if the dependency does not work as I suspect, we'll just have a double resource. The issues it will introduce are really minor: if svpc is true on both sides, destroy will fail but will complete successfully if relaunched.

morgante:
> Let me rework this as we will have a submodule anyway, and do a few tests. Then if the dependency does not work as I suspect, we'll just have a double resource.

SGTM, thanks!

PR #45, opened 2019-06-09
Second take at Shared VPC configuration.

ludoo:
As per the discussion in #42 this adds a simple submodule to configure Shared VPC network attachments, and the corresponding examples.

The module allows configuring VPC-level access through the `google_compute_shared_vpc_service_project` resource for 0-n projects, or subnet-level access for IAM-style members through the `google_compute_subnetwork_iam_binding` resource.

As discussed, there's a new output in the top-level module to expose the Shared VPC project id, which is used to trigger the inter-module dependency. Contrary to what I expected it seems to work, but even if it didn't it would not cause any major issue, as once SVPC has been turned on and applied, the submodule would work (meaning at most two consecutive apply runs).

There are no tests, as they would need extra service projects defined in the infrastructure used for the CI runs. I tested locally of course, and it works.

### DIFF
--- a/examples/submodule_svpc_access/README.md
+++ b/examples/submodule_svpc_access/README.md
@@ -6,7 +6,7 @@ The VPC has two subnets with no secondary ranges, service projects are configure
 
 - the first service project is granted VPC-level access
 - the second service project is granted subnet-level access to the second subnet
-- the this service project is granted subnet-level access to the first and second subnet
+- the third service project is granted subnet-level access to the first and second subnet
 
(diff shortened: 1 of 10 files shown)


examples/submodule_svpc_access/README.md
```diff
@@ -0,0 +1,25 @@
+# Shared VPC with service projects
+
+This simple example configures a shared VPC, and grants access to it to service projects.
+
+The VPC has two subnets with no secondary ranges, service projects are configured as follows:
+
+- the first service project is granted VPC-level access
+- the second service project is granted subnet-level access to the second subnet
+- the this service project is granted subnet-level access to the first and second subnet
```
aaron-lane:
```suggestion
- the third service project is granted subnet-level access to the first and second subnet
```

PR #46, opened 2019-06-13
Migrated to terraform v12. TO DO: some examples tune-up.

ivankorn:


ivankorn:
re-creating against migration branch

Issue #43, opened 2019-06-03
Add support for Terraform v0.12


aaron-lane:

btorresgil:
Error when using module v0.8.0 on Terraform v0.12:

```
$ terraform plan

Error: Unsupported block type

  on .terraform/modules/main.network_memory_store/terraform-google-modules-terraform-google-network-d1b6646/main.tf line 94, in resource "null_resource" "delete_default_internet_gateway_routes":
  94:   triggers {

Blocks of type "triggers" are not expected here. Did you mean to define
argument "triggers"? If so, use the equals sign to assign it a value.

```

skenmy:
@aaron-lane is this issue resolved by #48 ?

morgante:
Indeed it is.

PR #48, opened 2019-06-14
Migrated to terraform 0.12

ivankorn:
# 43: migrated to terraform 0.12
- switched to docker image with terraform 0.12
- migrated to new syntax
- defined type for some variables
- rewritten firewall's internal allow rules with dynamic block


aaron-lane:
Fixes #43.

PR #52, opened 2019-07-09
Fix terraform validate error

jeffmccune:
Without this patch, terraform validate with 0.12 is failing with the
following error message:

    Error: Unsupported block type

      on .terraform/modules/vpc/terraform-google-modules-terraform-google-network-d1b6646/main.tf line 94, in resource "null_resource" "delete_default_internet_gateway_routes":
      94:   triggers {

    Blocks of type "triggers" are not expected here. Did you mean to define
    argument "triggers"? If so, use the equals sign to assign it a value.

Note, this error also occurs when validating modules which depend on the
network module.  For example, this error has been copied from running
`make check_terraform` inside the project-factory module.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -91,7 +91,7 @@ resource "null_resource" "delete_default_internet_gateway_routes" {
     command = "${path.module}/scripts/delete-default-gateway-routes.sh ${var.project_id} ${var.network_name}"
   }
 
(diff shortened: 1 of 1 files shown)


jeffmccune:
Fix is already in the 0.12 release branch.

PR #47, opened 2019-06-14
Add support for Terraform v0.12

aaron-lane:

PR #55, opened 2019-07-09
Add a note about 0.11.x compatible version to README

Jberlinsky:
Per https://github.com/terraform-google-modules/terraform-google-network/pull/47#discussion_r300796223

We should probably add a note like this to each module's 0.12 upgrade PR.


Jberlinsky:
CI is green; I cancelled a spurious run.

PR #57, opened 2019-07-22
Added description and auto_create_subnetworks variables.

### LINKED ISSUE #56: Terraform Import of default network requires recreation on next apply
After importing the _default_ network, the next `terraform apply` requires destruction then recreation due to the following variables changing.
```bash
resource "google_compute_network" "network" {
      ~ auto_create_subnetworks         = true -> false # forces replacement
       ...
      - description                     = "Default network for the project" -> null # forces replacement
      ...
```

I would like to add these to the variables.tf so that they can be set in a user module.

kevensen:
Signed-off-by: Ken Evensen <kdevensen@google.com>

Fixes #56

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -63,6 +63,7 @@ variable "description" {
 }
 
 variable "auto_create_subnetworks" {
+  type        = bool
   description = "When set to true, the network is created in 'auto subnet mode' and it will create a subnet for each region automatically across the 10.128.0.0/9 address range. When set to false, the network is created in 'custom subnet mode' so the user can explicitly connect subnetwork resources."
-  default     = "false"
+  default     = false
 }
--- a/variables.tf
+++ b/variables.tf
@@ -54,3 +54,14 @@ variable "delete_default_internet_gateway_routes" {
   description = "If set, ensure that all routes within the network specified whose names begin with 'default-route' and with a next hop of 'default-internet-gateway' are deleted"
   default     = "false"
 }
+
+
+variable "description" {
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -54,3 +54,14 @@ variable "delete_default_internet_gateway_routes" {
...
   default     = "false"
 }
+
+
+variable "description" {
+  description = "An optional description of this resource. The resource must be recreated to modify this field."
+  default     = ""
+}
+
+variable "auto_create_subnetworks" {
```
morgante: This should be a boolean I believe.

variables.tf
```diff
@@ -54,3 +54,14 @@ variable "delete_default_internet_gateway_routes" {
   description = "If set, ensure that all routes within the network specified whose names begin with 'default-route' and with a next hop of 'default-internet-gateway' are deleted"
   default     = "false"
 }
+
+
+variable "description" {
```
morgante: Let's explicitly define the type.

variables.tf
kevensen: That was my immediate thought.  In main.tf, it is a string
https://github.com/terraform-google-modules/terraform-google-network/blob/4e7cd17b7c7e466fdec245b05d496cd80750a39a/main.tf#L20-L25

variables.tf
kevensen: will do

variables.tf
morgante: Historically, strings were auto-converted to booleans but now we can make explicit booleans.

kevensen:
@aaron-lane Done!

PR #58, opened 2019-07-25
Developer Experience 2.0 Updates

glarizza:
This PR represents the changes necessary to adapt the Network module to the new `cft/developer-tools` Docker image and the new testing workflow associated with the Developer Experience 2.0 update.


glarizza:
* [x] Remove unnecessary files within `test/` that have been consolidated into the developer-tools image
* [x] Update `Makefile` with make targets for `docker_run`, `docker_test_integration`, and `docker_test_lint`
* [ ] Add `Makefile.module` for module-specific make targets
* [x] Verify that integration tests execute successfully with `make docker_test_integration`
* [ ] Verify that lint tests execute successfully with `make docker_test_lint`

glarizza:
I've provided a checklist of tasks that need to be completed to completely migrate the network module to the new Developer Experience 2.0 workflow. Right now I've got integration tests working via `make docker_test_integration`, and there's a `make` target for `make docker_test_lint`, but `make test_lint` hasn't been wired up because the developer-tools image doesn't yet have a helper script for that action. Once the developer-tools image has been updated for lint testing, that change needs to be adapted to this module. Afterwards, Concourse needs to be updated to ensure the correct `make` targets are being executed and tests are passing.

glarizza:
Excluding CI changes for now until there's a decision on whether they are in-scope

Makefile.module
```diff
@@ -11,3 +11,8 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
morgante: We shouldn't need a custom Makefile just for this.

glarizza: Pulled it out

glarizza:
Unfortunately I have to transition to other work, but the checklist above is up to date. Currently both the integration and lint tests are working with the 0.0.1 (as of today) version of the developer-tools image. Concourse still remains using the old image and pipelines, which is why the pipelines are failing, but that work was on hold until there was a decision made on whether to update Concourse or just migrate directly to something else like Cloud Build.

morgante:
Approving and merging, we can update CI later.

PR #62, opened 2019-08-20
make secondary_ip_range optional

bmenasha:
now you don't need to supply values in the map when no secondary ranges should be created. we'll default to the empty list. 
also remove tabs.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -15,7 +15,7 @@
  */
 
 /******************************************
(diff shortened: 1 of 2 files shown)


morgante:
@bmenasha Thanks! This looks like a dupe of #16 though?

bmenasha:
yes. can be closed. thanks.

PR #72, opened 2019-09-19
Shared VPC access submodule improvements

ludoo:
* add management of host service agent role for GKE use cases
* improve documentation
* example fixes


morgante:
@skenmy Could you PTAL?

Issue #70, opened 2019-09-12
Updating any subnet in a list causes subsequent subnets to be re-created


onetwopunch:
## Expected behavior

When a subnet is modfied in source, only one resource is modified in the `plan`/`apply`

##  Actual behavior

When a subnet in the list of `var.subnets` is updated or deleted, it forces recreation of all subsequent subnets in the list.

## Example

If I were to update the `subnet_ip` for `subnet-01` it would force re-creation of `subnet-02` as well.

```
module "vpc" {
    source  = "terraform-google-modules/network/google"
    version = "~> 1.0.0"
  
   ...  
   
    subnets = [
        {
            subnet_name           = "subnet-01"
            subnet_ip             = "10.10.10.0/24"
            subnet_region         = "us-west1"
        },
        {
            subnet_name           = "subnet-02"
            subnet_ip             = "10.10.20.0/24"
            subnet_region         = "us-west1"
            subnet_private_access = "true"
            subnet_flow_logs      = "true"
            description           = "This subnet has a description"
        },
    ]
    ...
}
```

tfhartmann:
I also ran into this and started a feature branch that I got working. I should be able to open a PR later today 😄

PR #73, opened 2019-09-19
fix: updated subnets to use for_each rather then count

tfhartmann:
I also updated the outputs as well. This will be large state change however and probably needs to be a major release

This should address issue #70  - however I didn't update the routes resource which will likely have a similar problem


tfhartmann:
Hmm outputs have changed a bit, I'm going to see if I can create some tests for them

tfhartmann:
@skenmy @morgante  this PR should be ready for a review!  It will result in a pretty large state change for the end use, so some amount of state surgery will be required , something like `
```Shell
terraform state mv module.example.google_compute_subnetwork.subnetwork[0] module.example.google_compute_subnetwork.subnetwork[0]  module.example.google_compute_subnetwork.subnetwork[0] module.example.google_compute_subnetwork.subnetwork["us-east1/subnetwork1"]
```

examples/submodule_firewall/outputs.tf
```diff
@@ -28,3 +28,48 @@ output "admin_ranges" {
...
+  value       = module.test-vpc-module.network_self_link
+  description = "The URI of the VPC being created"
+}
+
+output "svpc_host_project_id" {
+  value       = module.test-vpc-module.svpc_host_project_id
+  description = "Shared VPC host project id."
+}
+
+output "subnets_names" {
```
ocervell: Output names plural_plural feels a bit weird. Any way we could consolidate those outputs (names, ips, regions, private_access, flow_logs, secondary_range into a map indexed by subnet name ?

Something like:
```
{
   subnet_1: {
      ip: 10.0.0.0/19
      region: us-east1
      private_access: true
      flow_logs: true
   },
   subnet_2: {
       ...
   }
}
```

outputs.tf
```diff
@@ -30,42 +30,42 @@ output "svpc_host_project_id" {
 }
 
 output "subnets_names" {
-  value       = google_compute_subnetwork.subnetwork.*.name
+  value       = [for network in google_compute_subnetwork.subnetwork : network.name] #google_compute_subnetwork.subnetwork[*].name
```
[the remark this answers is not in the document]
tfhartmann: whoops! Those should to totally have been removed! fixing!

examples/submodule_firewall/outputs.tf
tfhartmann: I like this idea a lot. For this PR I was mostly just trying not to break the outputs with the change.  Would you prefer me to take a swing at adding a new output to this PR, or do it in a separate one?

ludoo: +1

ludoo: +1 on changing the output format of the main module, I kind of fail to grasp the usefulness of adding subnet outputs to the firewall submodule.

tfhartmann: @ludoo yeah your right those aren't really needed in that example, I'll remove em

outputs.tf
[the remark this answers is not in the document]
aaron-lane: By the by, splat syntax is still valid in 0.12, just slightly modified:

```hcl
google_compute_subnetwork.subnetwork[*].name
```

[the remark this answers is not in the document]
tfhartmann: @aaron-lane true, do you have a preferred syntax?

examples/submodule_firewall/outputs.tf
tfhartmann: @ludoo ahh now I recall, if those outputs aren't defined the lint CI step will fail with these errors. 

`make docker_test_lint`

```
terraform_validate ./test/fixtures/submodule_firewall

Error: Reference to undeclared output value

  on outputs.tf line 33, in output "output_network_self_link":
  33:   value       = module.example.network_self_link

An output value with the name "network_self_link" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 38, in output "output_svpc_host_project_id":
  38:   value       = module.example.svpc_host_project_id

An output value with the name "svpc_host_project_id" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 43, in output "output_subnets_names":
  43:   value       = module.example.subnets_names

An output value with the name "subnets_names" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 48, in output "output_subnets_ips":
  48:   value       = module.example.subnets_ips

An output value with the name "subnets_ips" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 54, in output "output_subnets_regions":
  54:   value       = module.example.subnets_regions

An output value with the name "subnets_regions" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 59, in output "output_subnets_private_access":
  59:   value       = module.example.subnets_private_access

An output value with the name "subnets_private_access" has not been declared
in module.example.


Error: Reference to undeclared output value

  on outputs.tf line 64, in output "output_subnets_flow_logs":
  64:   value       = module.example.subnets_flow_logs

An output value with the name "subnets_flow_logs" has not been declared in
module.example.


Error: Reference to undeclared output value

  on outputs.tf line 69, in output "output_subnets_secondary_ranges":
  69:   value       = module.example.subnets_secondary_ranges

An output value with the name "subnets_secondary_ranges" has not been declared
in module.example.


Error: Reference to undeclared output value

  on outputs.tf line 74, in output "output_routes":
  74:   value       = module.example.routes

An output value with the name "routes" has not been declared in
module.example.

terraform_validate ./test/setup
Success! The configuration is valid.

make: *** [docker_test_lint] Error 1
```

outputs.tf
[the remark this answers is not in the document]
ocervell: @tfhartmann well, the new splat syntax might be better in that case, to avoid the for loop.

examples/submodule_firewall/outputs.tf
ocervell: @tfhartmann Just keep them like this and change the output format in a separate PR (or open an issue). @ludoo, thoughts ?

outputs.tf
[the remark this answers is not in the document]
tfhartmann: So I initially tried the spat syntax, and noticed tests failing. Which is why I created tests for outputs, apparently, the `splat` syntax, and the `for` syntax create a slightly different output, enough so that when I use the splat syntax  for the `subnets_ips` output,  the subnet firewall module isn't able to correctly create the resource.  Because of this, I'm going to recommend keeping with the `for` syntax so that `for_each` changes don't break  the outputs of the module.  I had thought that adding the output tests in  the secondary ranges example would have caught this, but it didn't seem to, maybe I'll add output tests to the firewall module as well. 

Here's the error that returns when using the Splat syntax 

```
  ×  gcloud: gcloud configuration (1 failed)
     ✔  Command: `gcloud compute firewall-rules describe submodule-firewall-timh-ingress-internal --project=dev-xpn-networking --format=json` exit_status should eq 0
     ✔  Command: `gcloud compute firewall-rules describe submodule-firewall-timh-ingress-internal --project=dev-xpn-networking --format=json` stderr should eq ""
     ×  Command: `gcloud compute firewall-rules describe submodule-firewall-timh-ingress-internal --project=dev-xpn-networking --format=json` internal rule should exist
     expected {"allowed" => [{"IPProtocol" => "icmp"}], "creationTimestamp" => "2019-09-26T06:53:50.896-07:00", "descript...etworking/global/firewalls/submodule-firewall-timh-ingress-internal", "sourceRanges" => ["0.0.0.0/0"]} to include {"sourceRanges" => ["10.10.20.0/24", "10.10.10.0/24"]}
     Diff:
     @@ -1,2 +1,14 @@
     -"sourceRanges" => ["10.10.20.0/24", "10.10.10.0/24"],
     +"allowed" => [{"IPProtocol"=>"icmp"}],
     +"creationTimestamp" => "2019-09-26T06:53:50.896-07:00",
     +"description" => "Allow ingress traffic from internal IP ranges",
     +"direction" => "INGRESS",
     +"disabled" => false,
     +"id" => "1123080907257904129",
     +"kind" => "compute#firewall",
     +"logConfig" => {"enable"=>false},
     +"name" => "submodule-firewall-timh-ingress-internal",
     +"network" => "https://www.googleapis.com/compute/v1/projects/dev-xpn-networking/global/networks/submodule-firewall-timh",
     +"priority" => 1000,
     +"selfLink" => "https://www.googleapis.com/compute/v1/projects/dev-xpn-networking/global/firewalls/submodule-firewall-timh-ingress-internal",
     +"sourceRanges" => ["0.0.0.0/0"],



Profile Summary: 0 successful controls, 1 control failure, 0 controls skipped
Test Summary: 2 successful, 1 failure, 0 skipped
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Verify failed on instance <submodule-firewall-local>.  Please see .kitchen/logs/submodule-firewall-local.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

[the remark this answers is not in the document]
tfhartmann: I added some tests for outputs in the firewall submodule, and it seems that when using the splat syntax the output for `subnet_ips`  is null, and doesn't exist, with the `for` syntax, it returns a list - `["10.10.10.0/24", "10.10.20.0/24"]`  

Here's the output of tests when using the splat syntx: 
`     ↺  Input 'output_subnets_ips' does not have a value. Skipping test.` 

It looks like tests that are skipped don't result in a fail  😢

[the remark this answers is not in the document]
tfhartmann: Here are the following outputs that have no value when using the splat syntax. ( when run against the `secondary_ranges` example. 

```Shell
     ↺  Input 'output_subnets_ips' does not have a value. Skipping test.
     ↺  Input 'output_subnets_flow_logs' does not have a value. Skipping test.
     ↺  Input 'output_subnets_names' does not have a value. Skipping test.
     ↺  Input 'output_subnets_private_access' does not have a value. Skipping test.
     ↺  Input 'output_subnets_regions' does not have a value. Skipping test.
     ↺  Input 'output_subnets_secondary_ranges' does not have a value. Skipping test.
```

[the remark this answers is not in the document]
tfhartmann: Sorry for the many comments 😄  Given that the output differs, I propose we stick with the `for` syntax since it matches the current output of the module.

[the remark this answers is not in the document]
morgante: Sticking with `for` sounds good to me.

examples/submodule_firewall/outputs.tf
morgante: Yes, let's keep this the same and not change output formats in this PR.

tfhartmann: Create this issue to track the above proposal about adding a new output map. https://github.com/terraform-google-modules/terraform-google-network/issues/77 . I'm also going to resolve this conversation. Please feel free to re-open if i've closed it in error.

outputs.tf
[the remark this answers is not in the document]
tfhartmann: Resolving this conversation. It's my understanding that we will use the `for` syntax for these outputs.

tfhartmann:
Please let me know if there are other requested changes, or concerns I can address in this PR 😃

main.tf
```diff
@@ -38,24 +45,16 @@ resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
...
-  secondary_ip_range       = [for i in range(length(contains(keys(var.secondary_ranges), var.subnets[count.index]["subnet_name"]) == true ? var.secondary_ranges[var.subnets[count.index]["subnet_name"]] : [])) : var.secondary_ranges[var.subnets[count.index]["subnet_name"]][i]]
-  description              = lookup(var.subnets[count.index], "description", null)
-}
-
-data "google_compute_subnetwork" "created_subnets" {
-  count   = length(var.subnets)
-  name    = element(google_compute_subnetwork.subnetwork.*.name, count.index)
-  region  = element(google_compute_subnetwork.subnetwork.*.region, count.index)
-  project = var.project_id
+  secondary_ip_range       = [for i in range(length(contains(keys(var.secondary_ranges), each.value.subnet_name) == true ? var.secondary_ranges[each.value.subnet_name] : [])) : var.secondary_ranges[each.value.subnet_name][i]]
```
ludoo: This is really hard to unravel, is there a way to make it more readable through whitespace, and possibly unnesting the condition by one level? Something like this (untested):

```
secondary_ip_range = [
    for i in range(
      contains(keys(var.secondary_ranges), each.value.subnet_name) == true
      ? length(var.secondary_ranges[each.value.subnet_name])
      : 0
    ) : var.secondary_ranges[each.value.subnet_name][i]
  ]
```

main.tf
```diff
@@ -38,24 +45,16 @@ resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
...
-  ip_cidr_range            = var.subnets[count.index]["subnet_ip"]
-  region                   = var.subnets[count.index]["subnet_region"]
-  private_ip_google_access = lookup(var.subnets[count.index], "subnet_private_access", "false")
-  enable_flow_logs         = lookup(var.subnets[count.index], "subnet_flow_logs", "false")
+  for_each                 = local.subnets
+  name                     = each.value.subnet_name
+  ip_cidr_range            = each.value.subnet_ip
+  region                   = each.value.subnet_region
+  private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
+  enable_flow_logs         = lookup(each.value, "subnet_flow_logs", "false")
```
ludoo: This is outside of the scope of this review so feel free to ignore, but I would very much like doing away with all the redundant prefixes in attribute names. Eg `subnet_name` should imho be just `name`, it's evident it refers to the subnet. Same for outputs like `network_name`, which should just be `name` as the module deals with a VPC and it's obvious what the output refers to.

outputs.tf
```diff
@@ -30,42 +30,41 @@ output "svpc_host_project_id" {
 }
 
 output "subnets_names" {
-  value       = google_compute_subnetwork.subnetwork.*.name
+  value       = [for network in google_compute_subnetwork.subnetwork : network.name]
```
ludoo: Why do you use a `for` here and in the other outputs, instead of the simpler splat expression `google_compute_subnetwork.subnetwork[*].name`? I find that `for` it more verbose and less readable (especially for people not yet familiar with 0.12 syntax).

main.tf
tfhartmann: Makes sense.  I updated it like this, which worked to keep all the tests happy. :) 

```Hcl
  secondary_ip_range = [
    for i in range(
      length(
        contains(
        keys(var.secondary_ranges), each.value.subnet_name) == true
        ? var.secondary_ranges[each.value.subnet_name]
        : []
    )) :
    var.secondary_ranges[each.value.subnet_name][i]
  ]
```

main.tf
tfhartmann: I figured I'd pick up work on #69 and since that work is already changing around the inputs I could remove the prefixes in the keys in that PR as well. Removing the `subnet_` prefix makes sense to me as well.

ludoo: Thanks! :)

outputs.tf
tfhartmann: Agreed, I had to use `for` because using the spat syntax in combination changed the behavior of the outputs and broke them, it's also why I added tests for the outputs.  When using the `for_each`  some of the outputs weren't correctly being created.  There's a resolved conversation somewhere in this PR, let me see if I can find that direct link!

tfhartmann: @ludoo here's the conversation! https://github.com/terraform-google-modules/terraform-google-network/pull/73#discussion_r328489995

tfhartmann:
> Thanks for the contribution. Just want to note that we need to write a migration guide before merging/releasing.

Is adding that to the README ok, or would you prefer it somewhere else?

morgante:
> Is adding that to the README ok, or would you prefer it somewhere else?

Preferable a separate doc + script, linked from changelog, following the [GKE example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CHANGELOG.md#v500---2019-09-25).

tfhartmann:
> 
> Preferable a separate doc + script, linked from changelog, following the [GKE example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CHANGELOG.md#v500---2019-09-25).

I rebased this branch against master, as well as attempted to add a migration doc, and a small helper script. 

I'm now fixing a bunch of small linting errors, after that I think this will be good to go

docs/upgrading_to_v2.0.md
```diff
@@ -0,0 +1,151 @@
...
+
+2.  Run the script to output the migration commands:
+
+    ```sh
+    $ MODULE_NAME="test-vpc-module" ./migrate.sh
+    terraform state mv module.example.module.test-vpc-module.google_compute_subnetwork.subnetwork[0] module.example.module.test-vpc-module.google_compute_subnetwork.subnetwork[\"us-west1/simple-project-timh-subnet-01\"]
+    terraform state mv module.example.module.test-vpc-module.google_compute_subnetwork.subnetwork[1] module.example.module.test-vpc-module.google_compute_subnetwork.subnetwork[\"us-west1/simple-project-timh-subnet-02\"]
+    ```
+
+3.  Execute the migration command
```
morgante: @tfhartmann We shouldn't require manually running commands: like the GKE module, the migration script should automatically run `terraform mv` for you (unless you pass the --dry-run flag).

tfhartmann: @morgante thats seems dangerous to me, but 🤷‍♂ I've updated the script and docs to behave in the way you've described.  The script will now take a `-d|--dry-run" flag, otherwise it'll migrate the state for you.

tfhartmann: *bump* @morgante  was there anything else you'd like to see in this PR?

tfhartmann:
Rebased against master this morning so that this PR wouldn't have merge conflicts

docs/upgrading_to_v2.0.md
```diff
@@ -0,0 +1,150 @@
...
+configuration and real physical resources that exist. As a result, no
+actions need to be performed.
+```
+
+### Migration Script
+
+1.  Download the script
+
+    ```sh
+    curl -O https://raw.githubusercontent.com/terraform-google-modules/terraform-google-network/master/scripts/migrate.sh
```
kopachevsky: Was not able to download the script with this command:
```
curl -O https://raw.githubusercontent.com/terraform-google-modules/terraform-google-network/master/scripts/migrate.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    15  100    15    0     0     40      0 --:--:-- --:--:-- --:--:--    41
➜  test  cat migrate.sh
404: Not Found
```

main.tf
```diff
@@ -38,24 +45,25 @@ resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
...
-  count   = length(var.subnets)
-  name    = element(google_compute_subnetwork.subnetwork.*.name, count.index)
-  region  = element(google_compute_subnetwork.subnetwork.*.region, count.index)
-  project = var.project_id
+  description              = lookup(each.value, "description", null)
+  secondary_ip_range = [
+    for i in range(
+      length(
+        contains(
+        keys(var.secondary_ranges), each.value.subnet_name) == true
```
kopachevsky: breaking  expression `contains keys(var.secondary_ranges), each.value.subnet_name) == true` on 2 lines made whole statement hardly readable from the first glance

docs/upgrading_to_v2.0.md
tfhartmann: > Was not able to download the script with this command:
> 
> ```
> curl -O https://raw.githubusercontent.com/terraform-google-modules/terraform-google-network/master/scripts/migrate.sh

Correct,  that URL won't work until after this PR has been merged. A current working example would be: 

`curl -O https://raw.githubusercontent.com/tfhartmann/terraform-google-network/tfhartmann-issue70/scripts/migrate.sh`

However that example will *stop* working after this PR is merged and I delete my feature branch. I wrote that step in anticipation of it being used after merge.

main.tf
tfhartmann: That feedback feels fairly subjective, if there is a documented best practice I'm happy to follow it.  I added the whitespace in response to this review.  https://github.com/terraform-google-modules/terraform-google-network/pull/73#discussion_r330389233 I don't honestly have a strong opinion on how this block should be format.

tfhartmann:
@morgante  I rebased this again this morning so as not to conflict with master.  I don't mean to nag,  could I get some feedback on this PR? Is there anything else you folks would like to see? I think I have addressed the outstanding feedback, if there is more I would be happy to address that as well.  Is this blocked for some reason?  This bug has help me up from using this module and I'd really like to use it.

tfhartmann:
> @tfhartmann Apologies for sitting on this so long. Code is looking great, but unfortunately the migration script isn't working for me:
Ah,  I think I see the problem, you actually have two modules named `vpc`  nested under `module.vpc-prod` and `module.vpc-nonprod` respectively.  Wrote the shell script with the input of MODULE_NAME so that people who were using multiple modules could explicitly determine which module they wanted to upgrade, and tested it against the examples, which use `module.generic.module.explicit` as a naming pattern, and against flat modules, - `module.name` both of which worked ok. I didn't test it against `module.explicit.module.generic` as a pattern. Honestly I think that may be a bit of an outlier in that I think most folks will use this against a module with a flat naming structure, however I've updated the script to just migrate *all* of the VPC modules it finds, which should now work against all the above combinations. 


> 
> ```shell
> $ MODULE_NAME=vpc ../../../modules/network/scripts/migrate.sh
> 
> Error: Invalid character
> 
>   on  line 2:
>   (source code not available)
> 
> Expected an attribute access or an index operator.
> ```
> 
> Dry-run looks incorrect too:
> 
> ```
> MODULE_NAME="vpc" ../../../modules/network/scripts/migrate.sh  --dry-run
> terraform state mv module.vpc-nonprod.module.vpc.data.google_compute_subnetwork.created_subnets[0] module.vpc-nonprod.module.vpc
> module.vpc-prod.module.vpc.google_compute_subnetwork.subnetwork[\"us-central1/clf-shared-net-gke\"]
> terraform state mv module.vpc-nonprod.module.vpc.google_compute_network.network module.vpc-nonprod.module.vpc
> module.vpc-prod.module.vpc.google_compute_subnetwork.subnetwork[\"clf-shared-net-nonprod\"]
> terraform state mv module.vpc-nonprod.module.vpc.google_compute_shared_vpc_host_project.shared_vpc_host[0] module.vpc-nonprod.module.vpc
> module.vpc-prod.module.vpc.google_compute_subnetwork.subnetwork[\"clf-shared-net-nonprod\"]
> terraform state mv module.vpc-nonprod.module.vpc.google_compute_subnetwork.subnetwork[0] module.vpc-nonprod.module.vpc
> module.vpc-prod.module.vpc.google_compute_subnetwork.subnetwork[\"us-central1/clf-shared-net-gke\"]
> terraform state mv module.vpc-prod.module.vpc.google_compute_subnetwork.subnetwork[0] module.vpc-nonprod.module.vpc
> ```
> 
> Would you mind taking another look at the migration script? I'd suggest simply reusing one of the existing [migration scripts](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/helpers/migrate.py) for consistency/reliability.

To be honest, when I looked at the python script, it looked like a bit more overhead then is actually needed. I would have to remove the bits that update the manifest files, and the parts that manipulate state looked like it was making calls to the shell. I think that rather than use python to invoke shell commands, it is simper to use a bash script in this instance. 

> 
> Also: could we place the migration script in `helpers/`? (Not `scripts/`.)

I moved the script to `helpers/`

morgante:
@tfhartmann Unfortunately I'm still having issues using the migration script, but I'm going to take a stab at fixing it myself.

PR #79, opened 2019-10-10
Add dynamic rules to Fabric firewall submodule

ludoo:
Aleks, this is a first shot at adding dynamic rule definitions to the firewall submodule, to make it a bit more flexible.

What I tried to implement here:

- enforcing structure for the rule definitions variable
- combine ingress and egress into a single variable/resource so that the module has few duplications
- correctly handle the incompatibility of tags/service accounts based source and targets

I did some minimal testing and it seems to work. This is an example variable for an ingress rule using both source and target tags, and a source range:

```
dynamic_rules = {
  dummy = {
    description          = "Dummy rule."
    direction            = "INGRESS"
    action               = "allow"
    ranges               = ["192.168.0.0"]
    sources              = ["spam-tag"]
    targets              = ["foo-tag", "egg-tag"]
    use_service_accounts = false
    rules = [
      {
        protocol = "tcp"
        ports    = []
      }
    ]
    extra_attributes = {}
  }
}
```

Can you give it a try and let me know your thoughts?

### DIFF
--- a/modules/fabric-net-firewall/.gitignore
+++ b/modules/fabric-net-firewall/.gitignore
@@ -1,0 +1,1 @@
+terraform.tfvars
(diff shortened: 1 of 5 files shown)


ludoo:
This has now been tested more thoroughly, outputs have been added, and the README file updated.

ludoo:
@morgante we need this for Fabric and it seems a minimal (and non-disruptive) enough implementation to go in after review, I would of course be interested in your comments.

Issue #65, opened 2019-09-03
local-exec is not picking up GOOGLE_APPLICATION_CREDENTIALS


komasoftware:
For my environment, there is a terraform specific service account with the right set of permissions to modify the infrastructure. So I generate a secrets.json key for the service account use this key as below to execute terraform :

`GOOGLE_APPLICATION_CREDENTIALS=secrets.json TF_VAR_xxxx=yyy TF_VAR_zzz=aaa apply --input=false`

This works just found, terraform is executing while impersonating the desired service account.
Except for `delete-default-gateway-routes.sh`
This script  is executing as local-exec and is not using the GOOGLE_APPLICATION_CREDENTIALS. The script is executing as the account setup by `gcloud auth application-default login`. 
This results is undesired and inconsistent behavioour IMO.

morgante:
As a workaround, you can also set `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=secrets.json` and it should work.

PR #83, opened 2019-10-14
Make use of GOOGLE_APPLICATION_CREDENTIALS in local-exec script.

omazin:
Fix for #65.

### DIFF
--- a/scripts/delete-default-gateway-routes.sh
+++ b/scripts/delete-default-gateway-routes.sh
@@ -36,6 +36,9 @@ function delete_internet_gateway_routes {
   done
 }
 
+if [ -n "${GOOGLE_APPLICATION_CREDENTIALS}" ]; then
+  gcloud auth activate-service-account --key-file="${GOOGLE_APPLICATION_CREDENTIALS}"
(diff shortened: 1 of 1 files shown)


omazin:
@morgante, thanks. can set the variable in the script `delete-default-gateway-routes.sh` as follows:
```bash
CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${GOOGLE_APPLICATION_CREDENTIALS}
``` 
but I don't think it's a good idea.

omazin:
Ok, found that `gcloud` generally does not use `GOOGLE_APPLICATION_CREDENTIALS` environment variable. The variable is used by GCP client libraries using the Application Default Credentials (ADC) strategy which automatically checks if the variable is set.
That's why `gcloud routes compute ...` commands in the local-exec script do not pick the `GOOGLE_APPLICATION_CREDENTIALS` variable, so in the PR we set the `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE` variable, which is respected by `gcloud`.

Issue #105, opened 2019-11-11
Enable Flow Logs deprecated


joshua9519:
Hi,
Not sure if this has been caught with another issue (I had a look but couldn't find anything - apologies if my search was lacking...)
The `enable_flow_logs` setting in Terraform has been deprecated in favour of `log_config` and therefore when using this module, you get a warning. 
If this hasn't been captured yet, I'd be happy to help with updating this. :+1: 
Thanks

morgante:
Thanks for the catch, looks like you're the first to notice!

I'd say this is what we should do:
1. Make a dynamic `log_config` block which is only added if `enable_flow_logs` is true.
2. Add a `flow_logs_interval` [variable](https://www.terraform.io/docs/providers/google/r/compute_subnetwork.html#aggregation_interval)
2. Add a `flow_logs_sampling` variable

We should base this work on the 2.0 branch: https://github.com/terraform-google-modules/terraform-google-network/pull/86

joshua9519:
Hi @morgante 
Thanks for the quick reply.
According to the warning:
```
Warning: "enable_flow_logs": [DEPRECATED] This field is being removed in favor of log_config. If log_config is present, flow logs are enabled.
```
So I think it should be more like a dynamic `log_config` and we deprecate `enable_flow_logs`?
What do you think?
Also, could the variable for `log_config` look like:
```
variable "log_config" {
  type = object({
    aggregation_interval = string
    flow_sampling = number
    metadata = string
  })
}
```
and remove the need for the other variables?
Let me know your thoughts :) 
Thanks 
Josh

morgante:
@joshua9519 I'd actually like to avoid changing the module interface. Internally, we can generate a dynamic `log_config` but keep the outside variables the same.

joshua9519:
Ah okay! I'm happy with that then, just thought I'd offer the suggestion. Would you like to keep `enable_flow_logs` too and use the logic you described above? My only issue with that is that the warning will still exist as we are still using that key.

morgante:
> Would you like to keep enable_flow_logs too and use the logic you described above? My only issue with that is that the warning will still exist as we are still using that key.

To be clear, we'd keep `enable_flow_logs` as a variable on the module but *not* set it on the resource. Instead, `enable_flow_logs` would drive the creation of the `log_config` block (and thus the warning would disappear).

joshua9519:
[...]

PR #108, opened 2019-11-14
Removing deprecated `enable_flow_logs` setting in favour of log_config

joshua9519:
When using this module with the latest Google provider, I noticed a warning about the deprecation of the `enable_flow_logs` setting in the `google_compute_subnetwork`. This was raised #105.  
This PR uses a dynamic block to define the `log_config` of a subnetwork if any of:
- `subnet_flow_logs_interval`
- `subnet_flow_logs_sampling`
- `subnet_flow_logs_metadata`  

are set. Or `subnet_flow_logs`is true.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -51,7 +51,7 @@ resource "google_compute_subnetwork" "subnetwork" {
   region                   = each.value.subnet_region
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
   dynamic "log_config" {
-    for_each = lookup(each.value, "subnet_flow_logs", false) || lookup(each.value, "subnet_flow_logs_interval", null) != null || lookup(each.value, "subnet_flow_logs_sampling", null) != null || lookup(each.value, "subnet_flow_logs_metadata", null) != null ? [{
+    for_each = lookup(each.value, "subnet_flow_logs", false) ? [{
       aggregation_interval = lookup(each.value, "subnet_flow_logs_interval", null)
       flow_sampling        = lookup(each.value, "subnet_flow_logs_sampling", null)
       metadata             = lookup(each.value, "subnet_flow_logs_metadata", null)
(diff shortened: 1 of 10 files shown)


main.tf
```diff
@@ -50,10 +50,21 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range            = each.value.subnet_ip
   region                   = each.value.subnet_region
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
-  enable_flow_logs         = lookup(each.value, "subnet_flow_logs", "false")
-  network                  = google_compute_network.network.name
-  project                  = var.project_id
-  description              = lookup(each.value, "description", null)
+  dynamic "log_config" {
+    for_each = lookup(each.value, "subnet_flow_logs", false) || lookup(each.value, "subnet_flow_logs_interval", null) != null || lookup(each.value, "subnet_flow_logs_sampling", null) != null || lookup(each.value, "subnet_flow_logs_metadata", null) != null ? [{
```
morgante: Instead of this complex conditional, I think we should probably just key off `subnet_flow_logs` (requiring it to be `true` if you want flow logging enabled).

test/integration/simple_project/controls/gcloud.rb
```diff
@@ -51,12 +55,51 @@
       end
     end
 
-    describe "enableFlowLogs" do
-      it "should be true" do
-        expect(data).to include(
-          "enableFlowLogs" => true
-        )
+    it "enableFlowLogs should be true" do
```
morgante: I believe the `enableFlowLogs` param is deprecated on the API, so we can probably drop this and simply test for the log config directly.

main.tf
joshua9519: So people would need to set `subnet_flow_logs` as well as the log config stuff?  
Or do you mean that `subnet_flow_logs` is no longer an option people can specify?  
Sorry for the confusion!

test/integration/simple_project/controls/gcloud.rb
joshua9519: No problem, will do!

main.tf
morgante: Right, `subnet_flow_logs` would control activation while the other ones control its behavior.

joshua9519:
Should I update them all?

morgante:
Yes, since we're updating the minimum required version with this change.

PR #117, opened 2019-11-26
Add subnet_log_config

bamaralf:
1) Since the "enable_flow_logs" var is deprecated in favor of "log_config" block, this PR adds the "subnet_log_config" var that receives a map and parses it to the "log_config" .
This PR also updates the examples with this change.
2) In the file main.tf of this module, this PR applies a fix the lint on the google_compute_route.route resource.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -55,12 +55,21 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range            = var.subnets[count.index]["subnet_ip"]
   region                   = var.subnets[count.index]["subnet_region"]
   private_ip_google_access = lookup(var.subnets[count.index], "subnet_private_access", "false")
-  enable_flow_logs         = lookup(var.subnets[count.index], "subnet_flow_logs", "false")
   network                  = local.network_self_link
   project                  = var.project_id
   secondary_ip_range       = [for i in range(length(contains(keys(var.secondary_ranges), var.subnets[count.index]["subnet_name"]) == true ? var.secondary_ranges[var.subnets[count.index]["subnet_name"]] : [])) : var.secondary_ranges[var.subnets[count.index]["subnet_name"]][i]]
   description              = lookup(var.subnets[count.index], "description", null)
   depends_on               = [google_compute_network.network]
+
+  dynamic "log_config" {
+    for_each = var.subnet_log_config
+
+    content {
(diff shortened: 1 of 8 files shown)


morgante:
Thanks for working on this. We actually have already fixed this in #108 for the upcoming 2.0 release.

variables.tf
```diff
@@ -45,6 +45,12 @@ variable "subnets" {
   description = "The list of subnets being created"
 }
 
+variable "subnet_log_config" {
+  type        = "map"
```
aaron-lane: Because this variable has an expected structure, it should be defined as an object rather than a map.

```suggestion
  type        = object({ aggregation_interval = string, flow_sampling = number, metadata = string })
```

main.tf
```diff
@@ -55,12 +55,21 @@ resource "google_compute_subnetwork" "subnetwork" {
...
   project                  = var.project_id
   secondary_ip_range       = [for i in range(length(contains(keys(var.secondary_ranges), var.subnets[count.index]["subnet_name"]) == true ? var.secondary_ranges[var.subnets[count.index]["subnet_name"]] : [])) : var.secondary_ranges[var.subnets[count.index]["subnet_name"]][i]]
   description              = lookup(var.subnets[count.index], "description", null)
   depends_on               = [google_compute_network.network]
+
+  dynamic "log_config" {
+    for_each = var.subnet_log_config
+
+    content {
+      aggregation_interval = lookup(log_config.value, "aggregation_interval", null)
```
aaron-lane: Based on the requested variable change, these expressions can be simplified to just access attributes.

variables.tf
```diff
@@ -45,6 +45,12 @@ variable "subnets" {
...
+  description = "If a value is set, flow logging for this subnetwork will be enabled."
```
aaron-lane:
```suggestion
  description = "Logging options for the subnetwork flow logs."
```

variables.tf
```diff
@@ -45,6 +45,12 @@ variable "subnets" {
...
+  default     = {}
```
aaron-lane:
```suggestion

  default     = {
    aggregation_interval = null
    flow_sampling        = null
    metadata             = null
  }
```

aaron-lane:
Whoops, ignore me. Missed #108.

PR #120, opened 2019-12-04
Pinned TF version and limited providers versions

### LINKED ISSUE #118: Required providers are not configured
Hello, in the **main.tf** file seems to be a deprecated option on **subnetwork** config called "enable_flow_logs". It seems to be replaced for log_config.

When I tried to execute a terraform plan using this module, it is returning this message:

**_Error: "enable_flow_logs": [REMOVED] This field is being removed in favor of log_config. If log_config is present, flow logs are enabled. Please remove this field_**

If is possible, could you remove this field from main.tf file?

Thanks in advance.

Best regards.

paulpalamarchuk:
* Fix #118


examples/delete_default_gateway_routes/main.tf
```diff
@@ -14,14 +14,6 @@
  * limitations under the License.
  */
 
-provider "google" {
```
morgante: @paulpalamarchuk Why are you removing these? We specifically do want them.

paulpalamarchuk: Moved to the `versions.tf`.

morgante:
@paulpalamarchuk Please rebase.

paulpalamarchuk:
@morgante Do we need exact "2.19.x" version of the google provider? Or I can limit it to `3> && >=2.19`?

examples/delete_default_gateway_routes/main.tf
morgante: I have no idea why you removed these. They're exactly right as they are. Please do not change anything about this line, this is exactly how provider versions should be declared in examples.

test/setup/versions.tf
```diff
@@ -1,5 +1,5 @@
 /**
```
morgante: Do not change this pinning approach. It is how we should do pinning within `test/setup`.

morgante:
Superseded by #130.

PR #129, opened 2019-12-11
add subnets output

adrian-gierakowski:
which is a map with keys `"${subnet_region}/${subnet_name}"` and values being the outputs
of the `google_compute_subnetwork` resource

with this output other modules can obtain data such as `self_link` for particular
subnets without running into resource recreation problems etc. when the number/order
of the subnets changes


adrian-gierakowski:
NOTE: I can add a description to the output if the PR is accepted

adrian-gierakowski:
@morgante done

PR #134, opened 2019-12-22
Provider update to 3.3.0

### LINKED ISSUE #128: Bump terraform-provider-google to 3.1.0
Hi People, 
I was wondering if there's a possibility to bump the `terraform-provider-google` version to the latest `3.1.0`, since there are quite some bugs that are fixed with it. The version is pinned to `2.19.0` and this is preventing me from using the network module in the same state. The only options for now I have are to use another state,  fork, drop or bump it via a PR.
If you agree with bumping the version I can also open a PR against the module with it.
Thanks!

### LINKED ISSUE #133: Run tests in parallel on CI
We should consider splitting tests as currently they run one-by-one and take some time.

Example in GKE repo:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/build/int.cloudbuild.yaml

bharathkkb:
fixes #128 
fixes #133 
- ```enable_flow_logs``` is removed, hence use flow logs config object with recommended defaults if subnet_flow_logs is true and subnet_flow_logs_interval/sampling/metadata is not provided
- updated and added new tests for the above
- fixed issue where next_hop_internet was being set to empty string and hence triggering ```only one of next_hop_gateway,next_hop_ilb,next_hop_instance,next_hop_ip,next_hop_vpn_tunnel can be specified, but next_hop_gateway,next_hop_instance were specified``` error

### DIFF
--- a/modules/subnets/main.tf
+++ b/modules/subnets/main.tf
@@ -33,9 +33,9 @@ resource "google_compute_subnetwork" "subnetwork" {
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
   dynamic "log_config" {
     for_each = lookup(each.value, "subnet_flow_logs", false) ? [{
-      aggregation_interval = lookup(each.value, "subnet_flow_logs_interval", null)
-      flow_sampling        = lookup(each.value, "subnet_flow_logs_sampling", null)
-      metadata             = lookup(each.value, "subnet_flow_logs_metadata", null)
+      aggregation_interval = lookup(each.value, "subnet_flow_logs_interval", "INTERVAL_5_SEC")
+      flow_sampling        = lookup(each.value, "subnet_flow_logs_sampling", "0.5")
+      metadata             = lookup(each.value, "subnet_flow_logs_metadata", "INCLUDE_ALL_METADATA")
     }] : []
     content {
       aggregation_interval = log_config.value.aggregation_interval
--- a/outputs.tf
+++ b/outputs.tf
@@ -55,8 +55,8 @@ output "subnets_private_access" {
 }
 
 output "subnets_flow_logs" {
-  value       = [for network in module.subnets.subnets : network.enable_flow_logs]
-  description = "Whether the subnets will have VPC flow logs enabled"
+  value       = [for network in module.subnets.subnets : network.log_config]
+  description = "If the subnets  have VPC flow logs enabled, config object is returned, else empty array"
 }
 
 output "subnets_secondary_ranges" {
--- a/outputs.tf
+++ b/outputs.tf
@@ -55,8 +55,8 @@ output "subnets_private_access" {
 }
 
 output "subnets_flow_logs" {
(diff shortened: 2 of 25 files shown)


outputs.tf
```diff
@@ -55,8 +55,8 @@ output "subnets_private_access" {
 }
 
 output "subnets_flow_logs" {
-  value       = [for network in module.subnets.subnets : network.enable_flow_logs]
-  description = "Whether the subnets will have VPC flow logs enabled"
+  value       = [for network in module.subnets.subnets : network.log_config]
```
morgante: This is a breaking change. Instead I think we should continue to return binary true/false of whether it's active.

modules/subnets/main.tf
```diff
@@ -33,9 +33,9 @@ resource "google_compute_subnetwork" "subnetwork" {
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
   dynamic "log_config" {
     for_each = lookup(each.value, "subnet_flow_logs", false) ? [{
-      aggregation_interval = lookup(each.value, "subnet_flow_logs_interval", null)
-      flow_sampling        = lookup(each.value, "subnet_flow_logs_sampling", null)
-      metadata             = lookup(each.value, "subnet_flow_logs_metadata", null)
+      aggregation_interval = lookup(each.value, "subnet_flow_logs_interval", "INTERVAL_5_SEC")
```
morgante: What's the reason for making this breaking change? I don't see a strong reason to enforce our own defaults.

outputs.tf
bharathkkb: Since 3.x.x removed enable_flow_logs arg, to keep the same interface I think we can check if log config object exists set true, if empty list set false. Does this approach work?

morgante:
Yes that works.
-- 

*Morgante Pell*

Strategic Cloud Engineer

morgantep@google.com

+1 212-565-7834

modules/subnets/main.tf
bharathkkb: In 3.x.x within log_config block at least one of aggregation_interval, flow_sampling, or metadata should be set. Hence if subnet_flow_logs is true, I believe the current dynamic block was setting empty blocks which threw at least one aggregation_interval, flow_sampling, or metadata should be set error. These defaults I propose are the same defaults as the provider docs.
[3.0.0 Subnetwork Upgrade Guide](https://www.terraform.io/docs/providers/google/guides/version_3_upgrade.html#resource-google_compute_subnetwork)
[Log Config Defaults](https://www.terraform.io/docs/providers/google/r/compute_subnetwork.html#log_config)

morgante: Yes, that should work.

outputs.tf
bharathkkb: @morgante I have switched back to returning binary t/f

modules/subnets/main.tf
bharathkkb: sg, so I am leaving these defaults as is.

PR #152, opened 2020-01-31
fix: service_project_ids type

marshallford:
Quoted type constraints are deprecated.

```
Warning: Quoted type constraints are deprecated

  on .terraform/modules/-.network_fabric-net-svpc-access/terraform-google-modules-terraform-google-network-270efc6/modules/fabric-net-svpc-access/variables.tf line 30, in variable "service_project_ids":
  30:   type        = "list"

Terraform 0.11 and earlier required type constraints to be given in quotes,
but that form is now deprecated and will be removed in a future version of
Terraform. To silence this warning, remove the quotes around "list" and write
list(string) instead to explicitly indicate that the list elements are
strings.
```


marshallford:
I'd love to know what the failing trigger's output is. Thanks!

marshallford:
Ah! My bad, tried to get clever and use the GitHub UI to create the PR. Fixed.

PR #155, opened 2020-02-11
Remove the code comment to use enable_logging

rsdomingues:
There is a commented line over the enable logging, just removing the comment and allowing the variable to be used.


modules/fabric-net-firewall/main.tf
```diff
@@ -135,7 +135,7 @@ resource "google_compute_firewall" "custom" {
   target_service_accounts = each.value.use_service_accounts ? each.value.targets : null
   disabled                = lookup(each.value.extra_attributes, "disabled", false)
   priority                = lookup(each.value.extra_attributes, "priority", 1000)
-  # enable_logging          = lookup(each.value.extra_attributes, "enable_logging", false)
+  enable_logging          = lookup(each.value.extra_attributes, "enable_logging", false)
```
morgante:
```suggestion
  enable_logging          = lookup(each.value.extra_attributes, "enable_logging", null)
```

PR #165, opened 2020-03-18
Adding type constraints to variables.

lpena-recurly:
For best practices and for terragrunt users [0]. Type constrains should be added [1]. Without specifying these constraints, applying a terragrunt plan/apply results in type errors [2], depending on the type constraints.

In terragrunt, "values are being passed in with environment variables and json, the type information is lost when crossing the boundary between Terragrunt and Terraform. You must specify the proper type constraint on the variable in Terraform in order for Terraform to process the inputs to the right type" [3]. Similar PR [4]

tl;dr lists and maps currently do not work with terragrunt because type constraints are not explicitly set.

[0] https://github.com/gruntwork-io/terragrunt
[1] https://www.terraform.io/docs/configuration/variables.html#type-constraints
[2] Invalid value for "inputMap" parameter: lookup() requires a map as the first
argument.
[3] https://terragrunt.gruntwork.io/docs/reference/config-blocks-and-attributes/#inputs
[4] https://github.com/terraform-google-modules/terraform-google-cloud-storage/pull/43

### DIFF
--- a/modules/fabric-net-firewall/variables.tf
+++ b/modules/fabric-net-firewall/variables.tf
@@ -16,24 +16,29 @@
 
 variable "network" {
   description = "Name of the network this set of firewall rules applies to."
+  type        = string
 }
 
 variable "project_id" {
   description = "Project id of the project that holds the network."
+  type        = string
 }
 
 variable "internal_ranges_enabled" {
   description = "Create rules for intra-VPC ranges."
+  type        = bool
   default     = false
 }
 
 variable "internal_ranges" {
   description = "IP CIDR ranges for intra-VPC rules."
+  type        = list(string)
   default     = []
 }
 
 variable "internal_allow" {
   description = "Allow rules for internal ranges."
+  type        = list(string)
   default = [
     {
       protocol = "icmp"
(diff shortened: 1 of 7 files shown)


modules/fabric-net-firewall/variables.tf
```diff
@@ -16,24 +16,29 @@
...
 
 variable "internal_ranges" {
   description = "IP CIDR ranges for intra-VPC rules."
+  type        = list(string)
   default     = []
 }
 
 variable "internal_allow" {
   description = "Allow rules for internal ranges."
+  type        = list(string)
```
morgante: This is a list of objects actually.

lpena-recurly: Great! got the fixes in and generated docs.

lpena-recurly: since one parameter is optional, there is no way to work around that with only 1 defined parameter. Removing based off of hashicorp/terraform#19898 . This will not affect the functionality of the module.

lpena-recurly: since one parameter is optional, there is no way to work around that with only 1 defined parameter. Removing based off of hashicorp/terraform#19898 . This will not affect the functionality of the module.

morgante:
@lpena-recurly Looks like tests are still failing. Could you get linters passing?

modules/fabric-net-firewall/variables.tf
```diff
@@ -16,53 +16,59 @@
...
 
 variable "internal_allow" {
   description = "Allow rules for internal ranges."
-  default = [
-    {
-      protocol = "icmp"
-    },
-  ]
+  type        = list(object({ protocol = string, ports = list(string) }))
+  default     = []
```
morgante: Please don't change the defaults.

lpena-recurly: got it. will resolve this soon. The linters would not pass because I changed the default. But the current default wont allow terragrunt to work. So I am looking for another solution.

morgante:
@lpena-recurly Are you still planning to work on this?

modules/fabric-net-firewall/variables.tf
forest: I had the same issue, and used this as the default. It does change the default but makes terragrunt happy.

```
default = [
    {
      protocol = "icmp"
      ports    = []
    },
  ]
```

I'm not sure what the correct solution is.

morgante: Does Terragrunt work with `type = list(any)`?

lpena-recurly:
@morgante Apologies for the delay. yes I will be fixing this soon.

morgante:
I'm closing this. @lpena-recurly please reopen if/when you can fix conflicts.

PR #167, opened 2020-03-31
Add variables type as first parameter on all variables

thiagonache:
It's a good practice and terragrunt does not work properly without it.


morgante:
CI failure is unrelated.

PR #169, opened 2020-04-06
add network output

umairidris:
Currently there is no way to get the network ID. Add the entire network as an output so users can use it however they need.


umairidris:
Thanks Morgante! I see "2 successful checks". Anything else needed here?

PR #174, opened 2020-04-16
helpers: Change attributes to attributes_flat

virta-jasonmay:
When running the migrate script, it encountered a KeyError `attributes` not found. I looked at the `terraform state pull` output, and saw there was an `attributes_flat` that likely changed in newer Terraform versions. I'm supplying this change to reflect that API change.

### DIFF
--- a/helpers/migrate.py
+++ b/helpers/migrate.py
@@ -304,7 +304,7 @@ class TerraformState:
         else:
             state_resource = state_resource_list[0]["instances"]
 
-        return state_resource[0]["attributes"][key]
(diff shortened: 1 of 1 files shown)


morgante:
Just to confirm, you're using Terraform 0.12?

virta-jasonmay:
Yes, specifically 0.12.24. If this worked with different versions of 0.12, should there be a check on version to determine the key name?

morgante:
Since it's just a migration script, I'm fine with updating to the latest behavior. Thanks for submitting a PR!

PR #180, opened 2020-04-29
Update main.tf

udayangadk:


udayangadk:
As google provider, it works with these fix

udayangadk:
As a google provider, it works only this fixing

morgante:
It looks like tests are failing.

PR #185, opened 2020-05-27
Fix/default routes

### LINKED ISSUE #175: Use google terraform provider to delete default routes rather than script
Now that `google_compute_network` supports deleting default routes with the parameter `delete_default_routes_on_create` can we replace the existing bash script / gcloud commands?

https://www.terraform.io/docs/providers/google/r/compute_network.html

ChandranshuRao14:
Fixes #175 
`var.delete_default_internet_gateway_routes` now uses `delete_default_routes_on_create` in the vpc module to delete default routes.

Happy to make any changes! Let me know your feedback.


variables.tf
```diff
@@ -51,8 +51,8 @@ variable "routes" {
   default     = []
 }
 
-variable "delete_default_internet_gateway_routes" {
-  description = "If set, ensure that all routes within the network specified whose names begin with 'default-route' and with a next hop of 'default-internet-gateway' are deleted"
+variable "delete_default_routes_on_create" {
```
bharathkkb, line 54: nit: I think we should add bool type and set as `false` since other variables seem to have it.  However this will be a breaking change, so I will defer to @morgante

modules/vpc/versions.tf
```diff
@@ -17,6 +17,6 @@
 terraform {
   required_version = "~> 0.12.6"
   required_providers {
-    google = "<4.0,>= 2.12"
+    google = "<4.0,>= 2.4"
```
bharathkkb, line 20: I see that `google_compute_network` added `delete_default_routes_on_create` with `2.4.0` but just wondering what was the reason for this since we already require a minimum of version of `2.12.0`.

variables.tf
morgante: Yes we should switch to boolean type. It's not a breaking change since strings can be cast to booleans.

bharathkkb: Nice, I assumed even with typecasting it would be technically considered a breaking change. TIL

modules/vpc/versions.tf
ChandranshuRao14: Mistake on my part. Thanks for catching that!

ChandranshuRao14:
> Since this provides the same functionality, is there any reason we need to change the variable name? I'd rather leave the module interface unchanged.

Makes sense! Reverted back to the original variable name and description to prevent breaking changes.

PR #187, opened 2020-05-28
fix: Add variable types for all variables

forest:
It's a good practice and terragrunt does not work properly without it.


morgante:
This looks like a duplicate of #165, I recommend looking there and making sure you've incorporated their feedback.

forest:
Yes. It is a duplicate. Sorry about that. Closing.

PR #191, opened 2020-06-29
Allow to pass target_tags to the default firewalls

rafaeljesus:
This PR adds the ability to specify the target_tags attribute for the default firewalls, rather than having them hardcoded within the code


rafaeljesus:
thank you, are you generating a new tag though?

PR #197, opened 2020-08-11
chore: relax tf version to allow 0.13 

### LINKED ISSUE #196: Support/Allow Terraform 0.13
Hi all - currently this module is semver-locked to Terraform 0.12.6, are we able to relax this to allow 0.13 or are there any known breaking changes? Happy to contribute/help out as needed.

bharathkkb:
fixes #196 
- relax version constraint to allow usage with 0.13
- Kitchen tests [run with `v0.13.0-rc1`](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/commit/88017ad3bc0d40473ac3d94546660328aeee878d) 
   - All tests passed
- module `for_each` spot checked, have not tested `depends_on`, but based on other modules I expect it to work

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -16,6 +16,7 @@
 driver:
   name: "terraform"
   command_timeout: 1800
+  verify_version: false
 
 provisioner:
   name: "terraform"
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 # Make will use bash instead of sh
 SHELL := /usr/bin/env bash
 
-DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 0.6.0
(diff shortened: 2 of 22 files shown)


.kitchen.yml
```diff
@@ -16,6 +16,7 @@
 driver:
   name: "terraform"
   command_timeout: 1800
+  verify_version: false
```
bharathkkb, line 19: @morgante currently int tests across all modules will fail with the latest `0.13` [dev tools image](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/763) without this due to https://github.com/newcontext-oss/kitchen-terraform/issues/412

morgante: I think we should probably add this line to all the modules.

Issue #199, opened 2020-08-11
Dedicated firewall rules module + adding to top level module


umairidris:
Currently the only CFT module to create firewall rules seems to be https://github.com/terraform-google-modules/terraform-google-network/tree/master/modules/fabric-net-firewall.

This seems to create a few default firewall rules and also needs to be a separate module call.

There is functionality to create custom rules. Would you be willing to accept a PR that pulls the custom rules logic out to its own submodule and adds it under the top level network module? Thus it would allow a user to create a network with custom firewall rules from one call.

morgante:
Yeah I think we need to refactor the firewall rules module a bit. Here's what I'd like to see:

1. Split the core rules logic into its own submodule
2. Put the default rules in a submodule (which calls the core rules module)

morgante:
Still needs to happen.

PR #200, opened 2020-08-14
feat: Add dedicated custom firewall rules module

umairidris:
#199


umairidris:
@morgante @bharathkkb my test seems to be hitting `Step #18 - "converge delete-default-gateway-routes-local":        Error: Error waiting to create Network: Error waiting for Creating Network: Quota 'NETWORKS' exceeded.  Limit: 5.0 globally.`

Any recommendation on this? The test seems unrelated to mine, but seems like there might be too many networks being created in the project. Should I merge the firewalls test with another test?

bharathkkb:
Hi @umairidris 
That seems right. I think merging tests are fine for this use case.
Another approach we did for GKE is spin up multiple projects in setup. https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/315/commits/b06c42541ca2fd830332635f58d3bb5818db77e3

umairidris:
Thanks Bharath. I'll go with the merge approach for now.

umairidris:
So I added the checks under secondary ranges, but `verify simple-project-with-regional-network-local` seems to be failing, which wasn't affected by this PR. Is it a known issue? 

`>>>>>>     Verify failed on instance <simple-project-with-regional-network-local>.  Please see .kitchen/logs/simple-project-with-regional-network-local.log for more details`

bharathkkb:
@umairidris that's odd last commit to main seems green, I will take a look.

modules/firewall-rules/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+}
+
+variable "network_name" {
+  description = "Name of the network this set of firewall rules applies to."
+}
+
+variable "rules" {
+  description = "List of custom rule definitions (refer to variables file for syntax)."
+  default     = []
+  type = list(object({
```
bharathkkb, line 28: I see that we require `object` here but in the main module take in `any` and convert into this list of objects for this submodule? Is there a reason why we just take in `any` here and and do the defaults lookup here?

I am thinking if users want to use this submod outside of main module, they will need to define everything since its a list of objects.

bharathkkb:
@umairidris looks like we have not pinned the CI images yet maybe thats why. Could you try pinning both images in lint and int CB yamls to `0.12.0`? Example: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/625/commits/e840f962582b34bd33603c2b1ca18ab708bbc763

modules/firewall-rules/variables.tf
umairidris: Ideally we should use list of objects everywhere. However as you pointed out users will need to define all variables possible through this. I chose to make the submodule list of objects to increase type checking while temporarily keeping the main module as type `any` until Terraform supports object defaults. 

I thought that was a decent compromise (if someone wants type checking they can use submodule but if they want a simpler interface then the main module implements that). Also, I was going to implement both as list of objects, but then the firewall rules would have a very different interface compared to the other fields such as subnets.

I am ok with whatever you guys suggest. If you want `any` for both levels then I can make the change.

umairidris:
Success! Thanks @bharathkkb. I still saw the failure for the test come up once more, but running /gcbrun to retry seems to have passed it, so it seems that specific test is a bit flaky for whatever reason.

modules/firewall-rules/variables.tf
bharathkkb: On further thought, I think if users want to ignore type checking they can just wrap the fw submodule like we do with main module. On a side note, I don't think defaults are slated till `0.14.0` 😞

modules/firewall-rules/main.tf
```diff
@@ -0,0 +1,47 @@
...
+  direction               = each.value.direction
+  network                 = var.network_name
+  project                 = var.project_id
+  source_ranges           = each.value.direction == "INGRESS" ? each.value.ranges : null
+  destination_ranges      = each.value.direction == "EGRESS" ? each.value.ranges : null
+  source_tags             = each.value.source_tags
+  source_service_accounts = each.value.source_service_accounts
+  target_tags             = each.value.target_tags
+  target_service_accounts = each.value.target_service_accounts
+  priority                = each.value.priority
```
bharathkkb, line 30: I saw `log_config` being passed through, but dont see it being set here?

umairidris: Good catch! This was an oversight on my part. I will fix this once I am done with the service identity features.

umairidris: Fixed.

morgante:
@umairidris Sorry we left this sitting for a while, do you mind resolving conflicts so we can merge?

comment-bot-dev:

umairidris:
@morgante done.

morgante (approved):
Thanks for your patience. 💯

modules/firewall-rules/versions.tf
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+terraform {
+  required_version = ">=0.12.6, <0.14"
```
jmymy, line 18: is this is a breaking change as version 3.0 didn't require <0.14 and now 3.1 does ?

jmymy: @morgante

jmymy:
fyi : https://github.com/terraform-google-modules/terraform-google-network/issues/247

modules/firewall-rules/versions.tf
jmymy: and i just saw your pr for the fix (https://github.com/terraform-google-modules/terraform-google-network/pull/245)  ignore me.

PR #202, opened 2020-08-25
0.13 upgrade

rhadley-recurly:


modules/vpc/override.tf
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+terraform {
+  required_version = ">= 0.13"
```
bharathkkb, line 18: We have relaxed constraints to allow usage with 0.13 with  https://github.com/terraform-google-modules/terraform-google-network/pull/197
We wont be dropping support for 0.12 immediately.

rhadley-recurly: Oh, I opened the PR to the wrong repo. I'm sorry. I meant to PR this to my fork.

But, to use 0.13 you also have to update your provider blocks and move them to override.tf.

rhadley-recurly: Reference: https://github.com/hashicorp/terraform/issues/25364

morgante: I don't see where it says required providers need to be in override.tf? We have already upgraded numerous other modules successfully using the minimal approach from #197.

rhadley-recurly: The issue comes up if you're using the module in a terraform base that already set required_providers. For example: we need to bring in our sops provider too. In 0.13 you can only have 1 required_providers block unless the additional block is in override.tf.

morgante: That's odd because my understanding was it's one block *per module*.

rhadley-recurly: This TF 0.13 PR here explains it: https://github.com/hashicorp/terraform/pull/24763

rhadley-recurly: Oh, interesting. Let me see if I can figure out why we're hitting this then. Maybe I'm doing something silly and don't need to fork. Thanks.

rhadley-recurly: Yes. The issue is that we are using terragrunt and terragrunt is copying an additional providers file in to the module.

Seriously, thanks for chatting with me on this! Much better to know the main problem than just fork and deal with it. Now I know what to actually fix.

morgante: Happy to help, thanks for investigating.

Issue #220, opened 2020-12-09
Support custom MTU when creating VPCs


memes:
When deploying it would be great if the module supported setting MTU on VPCs now that it has gone GA. Setting default to 1460 will preserve existing behaviour.

*NOTE:* This would change the Google provider dependency to [3.45.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v3.45.0).

bharathkkb:
fixed by #221

PR #221, opened 2020-12-09
Support custom MTU on VPCs

memes:
This is a potential update to address feature request #220. It adds an `mtu` variable to root and vpc modules that will set MTU on `google_compute_network` resource.

The update has been tested in a few non-production VPCs without any problems yet.

*Note:* MTU suport has a dependency on release [3.45.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v3.45.0) of google provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Then perform the following commands on the root folder:
 | auto\_create\_subnetworks | When set to true, the network is created in 'auto subnet mode' and it will create a subnet for each region automatically across the 10.128.0.0/9 address range. When set to false, the network is created in 'custom subnet mode' so the user can explicitly connect subnetwork resources. | `bool` | `false` | no |
 | delete\_default\_internet\_gateway\_routes | If set, ensure that all routes within the network specified whose names begin with 'default-route' and with a next hop of 'default-internet-gateway' are deleted | `bool` | `false` | no |
 | description | An optional description of this resource. The resource must be recreated to modify this field. | `string` | `""` | no |
+| mtu | The network MTU (default '1460'). Must be a value between 1460 and 1500 inclusive. | `number` | `1460` | no |
 | network\_name | The name of the network being created | `any` | n/a | yes |
 | project\_id | The ID of the project where this VPC will be created | `any` | n/a | yes |
 | routes | List of routes being created in this VPC | `list(map(string))` | `[]` | no |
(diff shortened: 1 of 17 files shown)


morgante (changes requested):
You'll also need to update the version in `examples/`.

memes:
I've changed the examples so that they pass with `make docker_test_integration` executed locally. Full tests failed until I made the change in 8063a17 too.

morgante (approved):
Thanks for the contribution. Test failures look unrelated.

jmymy:
Is this a breaking change? It's a destructive action to change the MTU on existing VPCs

`  # module.vpc.google_compute_network.network must be replaced
+/- resource "google_compute_network" "network" {
        auto_create_subnetworks         = false
        delete_default_routes_on_create = false
      + gateway_ipv4                    = (known after apply)
      ~ id                              = "XXXXX" -> (known after apply)
      ~ mtu                             = 0 -> 1460 # forces replacement
        name                            = "XXXXX"
        project                         = "XXXXXX"
        routing_mode                    = "GLOBAL"
      ~ self_link                       = "XXXXXX" -> (known after apply)
    }`

jmymy:
https://github.com/hashicorp/terraform-provider-google/blob/master/google/resource_compute_network.go#L78

memes:
Tested and confirmed that is is a breaking change for those upgrading from v2.x to v3.0.0. Personally I think this is acceptable given the module's major version has changed for this and other changes, but I'm open to other points of view.

FYI - By adding an explicit `mtu = 0` input to the module I avoided recreating a test network on upgrade from v2.6.0 to v3.0.0. @jmymy can you confirm that works in your environment too?

Options:
1. Leave default MTU variable at `1460` to match intent, and document the workaround
2. Set default input value for `mtu` to `0` or `null`; this will prevent breakage for those upgrading from v2.x.x to v3.0.x but will bring in the same problem for v3.0.0 users who do not explicitly set an MTU value when the next release is published.

@morgante @bharathkkb - how do you want to proceed?

bharathkkb:
A quick look at the API seems to imply that it is possible to update MTU. If that is the case this should not be marked as `ForceNew` in the provider. I will open an issue on the provider to confirm, but short term setting it to `null` could be a workaround.

morgante:
Indeed, you can override the MTU as a workaround. Since this is a (breaking) major version change, I don't think we need to change the module behavior.

Issue #201, opened 2020-08-19
Test verify simple-project-with-regional-network-local is flaky


umairidris:
See https://pantheon.corp.google.com/cloud-build/builds/373c3073-beda-4e8b-8af4-525d94701462;step=7?project=cloud-foundation-cicd for sample failure.

Retrying seems to fix it.

```
$$$$$$ Running command `terraform output -json` in directory /workspace/test/fixtures/simple_project_with_regional_network
inspec-gcp: Verifying
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Verify failed on instance <simple-project-with-regional-network-local>.  Please see .kitchen/logs/simple-project-with-regional-network-local.log for more details
```

PR #229, opened 2021-01-13
Fix/vpc peering #228

### LINKED ISSUE #228: network-peering module generate error when length of local and peered vpc names exceed 62 chars
[vpc-peering](https://github.com/terraform-google-modules/terraform-google-network/tree/master/modules/network-peering) module fails when the length of local and peered vpc names are over 62 chars. It will be great if we allow passing vpc peering name or check if the combined length is over 62 chars then crop the combination and add some random chars at the end of peering name. 

Current Code:

```
locals {
  local_network_name = element(reverse(split("/", var.local_network)), 0)
  peer_network_name  = element(reverse(split("/", var.peer_network)), 0)
}

resource "google_compute_network_peering" "local_network_peering" {
  provider             = google-beta
  name                 = "${var.prefix}-${local.local_network_name}-${local.peer_network_name}"
....
}

resource "google_compute_network_peering" "peer_network_peering" {
  provider             = google-beta
  name                 = "${var.prefix}-${local.peer_network_name}-${local.local_network_name}"
....
}
```

imrannayer:
Fixed #228

### DIFF
--- a/modules/network-peering/README.md
+++ b/modules/network-peering/README.md
@@ -51,8 +51,10 @@ module "peering-a-c" {
 | export\_local\_custom\_routes | Export custom routes to peer network from local network. | `bool` | `false` | no |
 | export\_peer\_custom\_routes | Export custom routes to local network from peer network. | `bool` | `false` | no |
 | local\_network | Resource link of the network to add a peering to. | `string` | n/a | yes |
+| local\_network\_peering\_name | Name of local network peering | `string` | `""` | no |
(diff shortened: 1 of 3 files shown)


morgante (changes requested):
I'd rather not complicate the interface with custom names, but the substring addition is good.

imrannayer:
I'll remove custom names

bharathkkb:
Tests are failing due to a similar error like https://github.com/terraform-google-modules/terraform-google-network/issues/201

morgante (approved):
Thanks for the contribution!

PR #236, opened 2021-02-02
fix: Set firewall logging for each rule

bgdanix:
With the current version, it is not possible to enable firewall logs. This PR fixes the issue by changing the firewall module to look at the `extra_attributes` inside the `custom_rules` var when dynamically building the `log_config`.
Variable types have also been updated (where it made sense) for fabric-net-firewall module.


comment-bot-dev:

morgante (approved):
Thanks for the contribution.

PR #272, opened 2021-04-01
Route example for docs

betsy-lichtenberg:


examples/routes/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "project_id" {
+  description = "The ID of the project where the routes will be created"
+}
+
+variable "network_name" {
```
morgante, line 21: Please don't include variables you're not using. In this case, the only variable you referenced is "project_id."

morgante (changes requested):

examples/routes/versions.tf
```diff
@@ -0,0 +1,29 @@
...
+  required_version = ">= 0.13.0"
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "<4.0,>= 2.12"
+    }
+  }
+
+  provider_meta "google" {
+    module_name = "blueprints/terraform/terraform-google-network:routes/v3.2.0"
```
morgante, line 27: This is not necessary.

examples/routes/variables.tf
betsy-lichtenberg: Fixed

examples/routes/versions.tf
betsy-lichtenberg: Fixed

morgante (approved):

PR #299, opened 2021-05-31
Adds `subnet_ids` as an output

samcre:
Just adds every `subnet_id` as an output. Useful for use with other modules that requires a list of the subnet ids.


comment-bot-dev:

bharathkkb (approved):
Thanks for the PR @samcre

PR #324, opened 2021-10-06
Remove deprecated test setup instructions

rsinnet:
In #58, the root Makefile was refactored and it seems the intention thereafter is to run the integration tests exclusively via the Docker container. Remove the deprecated test/setup/README.md which is now handled through a Make target.


rsinnet:
> Thanks for PR! Yes now we prefer everything being run in the container so we can actually remove this file altogether. Our authoritative source should be https://github.com/terraform-google-modules/terraform-google-network/blob/master/CONTRIBUTING.md

Ah that makes a lot of sense! Thanks for the explanation.

As I was reading over [CONTRIBUTING.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/CONTRIBUTING.md), I noticed there is a link to this [README.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/test/setup/README.md) there:
https://github.com/terraform-google-modules/terraform-google-network/blob/1cc50ca5c9ce5a1c33b30e4f24fa6b562688afd2/CONTRIBUTING.md#L39

It seems like this [README.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/test/setup/README.md) is intended to describe how to set up a sandboxed project in which to then safely run the integration tests (via Docker). I'm not 100% sure? Maybe that reference is obsolete? If not, I wonder if it would be better to remove the ending of the [README.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/test/setup/README.md) which describes how to run the tests (and perhaps point have it point the reader back to the root [README.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/README.md)). Another option would be to move the relevant bits into a section in [CONTRIBUTING.md](https://github.com/terraform-google-modules/terraform-google-network/blob/master/CONTRIBUTING.md) and then delete this file altogether.

Do you think either of those is a good option?

bharathkkb:
@rsinnet sorry for the delay!

> As I was reading over CONTRIBUTING.md, I noticed there is a link to this README.md there. It seems like this README.md is intended to describe how to set up a sandboxed project in which to then safely run the integration tests (via Docker). 

I believe the link is pointing to `test/setup` and not particularly to the readme. The intention for that is to point out the underlying tf configuration that will be used by `docker_test_prepare` so that users can inspect it before proceeding to create the sandbox project using `docker_test_prepare` helper. 

I think [this section](https://github.com/terraform-google-modules/terraform-google-network/blob/master/CONTRIBUTING.md#test-environment) of the contributing guide describes the needed setup so we can go ahead with removing the `test/setup/Readme.md`. Let me know if this is not sufficient or if you encountered any additional friction while setting this up.

rsinnet:
> 

After looking more closely, you're right. I renamed this PR and removed the README.md. I think that will prevent future developers from accidentally following the instructions. Good call.

rsinnet:
> @rsinnet sorry for the delay!
> 
> > As I was reading over CONTRIBUTING.md, I noticed there is a link to this README.md there. It seems like this README.md is intended to describe how to set up a sandboxed project in which to then safely run the integration tests (via Docker).
> 
> I believe the link is pointing to `test/setup` and not particularly to the readme. The intention for that is to point out the underlying tf configuration that will be used by `docker_test_prepare` so that users can inspect it before proceeding to create the sandbox project using `docker_test_prepare` helper.
> 
> I think [this section](https://github.com/terraform-google-modules/terraform-google-network/blob/master/CONTRIBUTING.md#test-environment) of the contributing guide describes the needed setup so we can go ahead with removing the `test/setup/Readme.md`. Let me know if this is not sufficient or if you encountered any additional friction while setting this up.

After looking more closely, you're right. I renamed this PR and removed the README.md. I think that will prevent future developers from accidentally following the instructions. Good call.

PR #335, opened 2021-11-10
Updates google provider constraints

jackwhelpton:


morgante (changes requested):
Please update the examples as well: https://github.com/terraform-google-modules/terraform-google-network/tree/master/examples

comment-bot-dev:

jackwhelpton:
> Please update the examples as well: https://github.com/terraform-google-modules/terraform-google-network/tree/master/examples

Makes sense. There's a mixture of constraints in main.tf for those... some are using ~> and some have only a minimum constraint, haven't seen any that define an explicit upper bound yet. Are you happy to make all of these `>= (current value)`?

morgante:
> Thinking aloud... best practice is to use ~> for root modules, so perhaps I should update them all to be ~> 4.0.0.

Let's use `~> 4.0` please.

jackwhelpton:
> Let's use `~> 4.0` please.

First stab at this done. I've assumed that anything beginning `submodule_` is meant to demonstrate how to use the modules within a module, whilst the others are straight Terraform projects, but that may be erroneous.

There was a mixture of different ways of applying provider constraints in these examples: some seemed best suited for modules and some for straight projects, so it was a bit confusing. If you want this laid out differently just let me know.

examples/basic_secondary_ranges/main.tf
```diff
@@ -14,16 +14,12 @@
  * limitations under the License.
  */
 
+provider "google" {
```
morgante, line 17: I believe this approach is deprecated, please switch back to `required_providers`.

morgante (changes requested):

examples/basic_secondary_ranges/main.tf
jackwhelpton: Gotcha: there was a mix of the two approaches in the existing examples. I'd assumed one was intended for modules and one for projects, but I'll switch them all to use `required_providers`.

jackwhelpton: I'm going to move the required providers into the `versions.tf` file where it exists.

jackwhelpton: Also, I remember this problem from my contribution to one of the other models: I can't retrieve the Cloud Build logs, although I can see the `int` and `lint` triggers have failed. If you can paste any info here so I know what to look at/fix that'd be good.

morgante:
@jackwhelpton You don't need access to the lint logs. The above comment is updated with the issues. Please continue address until it fixes (or simply run the linter locally).

jackwhelpton:
Daft question 94 (or whatever we're up to now): a lot of the remaining failures look like they come from examples that reference specific versions of the modules within this repo.

Given that we need to update this module in order to change the provider versions, how do we resolve this circular dependency?

```
module "google_compute_route" {
  source       = "terraform-google-modules/network/google//modules/routes"
  version      = "~> 3.2.0"
  project_id   = var.project_id # Replace this with your project ID in quotes
  network_name = "default"

  routes = [
    {
      name              = "egress-internet"
      description       = "route through IGW to access internet"
      destination_range = "0.0.0.0/0"
      tags              = "egress-inet"
      next_hop_internet = "true"
    }
  ]
}
```

morgante:
> Given that we need to update this module in order to change the provider versions, how do we resolve this circular dependency?

Good question. For now, can you just leave a comment on the examples which will need to be updated?

jackwhelpton:
> For now, can you just leave a comment on the examples which will need to be updated?

Sure, I just want to validate that's the only failure type left. When you say a comment, do you mean on the PR? In the code? If the latter, a comment on the version line for any modules pinned that way?

morgante:
> When you say a comment, do you mean on the PR? In the code? If the latter, a comment on the version line for any modules pinned that way?

Yep, a comment in the code (at the top of the file is fine).

> There's a mix of behavior here, some of the modules that are working are doing so because they use relative paths. To see the most pronounced discrepancy, note that simple_project_with_regional_network uses a relative path and succeeds, whilst simple_project uses a (pinned) absolute path and hence fails. Would using relative paths throughout be an idea? Less helpful if these examples are intended to be something an end user could copy into their project and use directly, though.

Relative paths is our *default* approach since it works for testing quite nicely.

However, some examples are [embedded directly](https://cloud.google.com/vpc/docs/using-routes#terraform) in our documentation. These ones need to use pinned paths.

jackwhelpton:
> Relative paths is our _default_ approach since it works for testing quite nicely.
> 
> However, some examples are [embedded directly](https://cloud.google.com/vpc/docs/using-routes#terraform) in our documentation. These ones need to use pinned paths.

Ah, that makes sense, and explains the mix. Is there a way to track which ones are embedded that way? Beyond "the ones with pinned paths", although I guess that might be the answer.

Commit to add those comments coming up.

morgante:
> Ah, that makes sense, and explains the mix. Is there a way to track which ones are embedded that way? Beyond "the ones with pinned paths", although I guess that might be the answer.

Bingo. :)

jackwhelpton:
Comments added.

examples/basic_vpc_peering/versions.tf
```diff
@@ -16,4 +16,13 @@
 
 terraform {
   required_version = ">=0.12.6"
+
+  required_providers {
+    google = {
+      version = "~> 4.0"
```
morgante, line 22: Ah, I think I was unclear. For the examples where the module is pinned, we should stick to `~> 3.x`.

morgante (changes requested):

examples/basic_vpc_peering/versions.tf
jackwhelpton: Makes sense, so the live examples still work... those updates were made before I was clear on that. I wonder if we'll still end up with a version clash, but I'll make that correction now and let's see.

jackwhelpton: Done.

morgante: Looks like linters are still failing. Please fix.

jackwhelpton: Isn't that an expected conflict between the versions required by the pinned and non-pinned versions?

I need to head out for the evening now, will take a further look tomorrow.

morgante: Nope, you'll notice it is failing on just `test/setup` now. Which you can actually leave completely untouched.

jackwhelpton: Great, that did it, thanks. Actually not _completely_ untouched, as it was also using the old form of representing constraints.

morgante (approved):
Thanks for your patience!

mitchelladam:
any news on when this might get released?

bharathkkb:
@mitchelladam we usually release every ~2 weeks, you can track the release here https://github.com/terraform-google-modules/terraform-google-network/pull/336

Issue #339, opened 2021-11-24
Support for role and purpose in network module


jawnsy:
### TL;DR

Only subnets-beta supports configuring purpose and role, but this seems to have been promoted to a non-beta feature, so we can add support to the top-level module

### Terraform Resources

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork

### Additional information

I think the proposed change consists of modifying subnets, and potentially removing the now-unnecessary subnets-beta module (since all fields are supported by the main subnets module). This change consists of adding the inputs:

* purpose
* role

for the subnets module: https://github.com/terraform-google-modules/terraform-google-network/blob/a7b91bc1a70c323b99eecc229a70b449b6bda401/modules/subnets/main.tf#L28-L33

jawnsy:
I think this is still a valid problem

bharathkkb:
@jawnsy I believe we added this via https://github.com/terraform-google-modules/terraform-google-network/pull/343

PR #340, opened 2021-11-24
Add role and purpose for subnet settings

jawnsy:
Closes: #339

### DIFF
--- a/modules/subnets/main.tf
+++ b/modules/subnets/main.tf
@@ -31,6 +31,8 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range            = each.value.subnet_ip
   region                   = each.value.subnet_region
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
+  purpose                  = lookup(each.value, "subnet_purpose", null)
+  role                     = lookup(each.value, "subnet_role", null)
   dynamic "log_config" {
     for_each = lookup(each.value, "subnet_flow_logs", false) ? [{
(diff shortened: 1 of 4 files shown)


modules/subnets/main.tf
```diff
@@ -31,6 +31,8 @@ resource "google_compute_subnetwork" "subnetwork" {
   ip_cidr_range            = each.value.subnet_ip
   region                   = each.value.subnet_region
   private_ip_google_access = lookup(each.value, "subnet_private_access", "false")
+  purpose                  = lookup(each.value, "subnet_purpose", null)
```
jawnsy, line 34: followed the `subnet_xxx` pattern here, but `subnet-beta` declares these as just `role` and `purpose` (no prefix)

comment-bot-dev:

bharathkkb (changes requested):
Thanks for the PR @jawnsy 
We will also need to update versions in [main module](https://github.com/terraform-google-modules/terraform-google-network/blob/master/versions.tf) and [subnets](https://github.com/terraform-google-modules/terraform-google-network/blob/a7b91bc1a70c323b99eecc229a70b449b6bda401/modules/subnets/versions.tf#L23) to minimum 3.83

modules/subnets/main.tf
bharathkkb: We can go with subnet prefix since others follow this pattern in this module

jawnsy:
@bharathkkb Thanks for the quick review and the pointers to the code that needed to be updated! I've bumped the minimum versions to 3.83

PR #433, opened 2023-03-16
feat(firewall_rules)!: allow destination_ranges and source_ranges to be specified simultaneously

fredvollmer:
Note: This PR relies on behavior that is currently in preview.

The `firewall_rules` module currently accepts only a single `ranges` input, which is used for one of `destination_ranges` or `source_ranges` depending upon the rule `type`.

This PR adds explicit `source_ranges` and `destination_ranges` inputs, which will allow rules to specify both a source and destination range set.

The `ranges` input has been marked as deprecated. However, the original behavior has been maintained for backwards compatibility.

## Background
For INGRESS rules, it is now possible to restrict incoming packets to those with a certain destination by specifying `destination_ranges`.

For EGRESS rules, the set of allowed sources can be further restricted by specifying `source_ranges`.

From the docs: 

(https://cloud.google.com/vpc/docs/firewalls#targets_and_ips_ingress)
> If the ingress firewall rule includes a destination IP address range, the packet's destination must fit within one of the explicitly defined destination IP address ranges

(https://cloud.google.com/vpc/docs/firewalls#targets_and_ips_egress)
> If the egress firewall rule includes source IP address ranges, the target VMs are still limited to the source IP addresses mentioned previously, but the source parameter can be used to refine that set 

BREAKING CHANGE: adding new properties to `rules` object will require caller to pass in those properties. Terraform provides optional types, but they are not supported in this repo's minimum TF version of 12.6.


apeabody:
Thanks for the contribution @fredvollmer!

> BREAKING CHANGE: adding new properties to `rules` object will require caller to pass in those properties. Terraform provides optional types, but they are not supported in this repo's minimum TF version of 12.6.

Looks like the doc probably need to be bumped as the root and sub-module in question is requiring TF >= 0.13.  As this is already a breaking change, what version would be ideal?

https://github.com/terraform-google-modules/terraform-google-network/blob/master/versions.tf#L18

apeabody:
Hi @fredvollmer - From the LINT:

```
Running terraform fmt
modules/firewall-rules/variables.tf
--- old/modules/firewall-rules/variables.tf
+++ new/modules/firewall-rules/variables.tf
@@ -28,10 +28,10 @@
   description = "List of custom rule definitions (refer to variables file for syntax)."
   default     = []
   type = list(object({
-    name                    = string
-    description             = string
-    direction               = string
-    priority                = number
+    name        = string
+    description = string
+    direction   = string
+    priority    = number
     # Deprecated; use source_ranges or destination_ranges instead.
     # For rule of type INGRESS, will be used as source_ranges if not defined.
     # For rule of type EGRESS, will be used as destination_ranges if not defined.
Error: terraform fmt failed with exit code 3
Check the output for diffs and correct using terraform fmt <dir>
```

PR #443, opened 2023-04-25
fix: update compute instance to use reserved ip

msampathkumar:
As you can understand from `compute_shared_instance_with_reserved_ip_create`, this compute instance was supposed to created with reserved ip address.

fixes: b/275008920


examples/basic_shared_vpc/main.tf
```diff
@@ -57,7 +57,7 @@ resource "google_compute_address" "internal" {
 resource "google_compute_instance" "reserved_ip" {
   project      = var.service_project
   zone         = "us-central1-a"
-  name         = google_compute_address.internal.self_link
+  name         = "reserved-ip"
```
bharathkkb, line 60: Just to make to clearer that this is a VM resource.

```suggestion
  name         = "reserved-ip-instance"
```

bharathkkb (approved):

PR #458, opened 2023-06-14
feat: add destination_ranges and source_ranges in firewall rules

### LINKED ISSUE #461: Firewall rules for terraform-google-network module
### TL;DR

Hi there, one of our enterprise customers currently uses terraform-google-modules/terraform-google-network in their IaC pipeline, and they would like to have source and destination to be used in egress and ingress firewall rules. 

### Terraform Resources

_No response_

### Detailed design

_No response_

### Additional information

just FYI: Below is the comment from our customer: 
***
Seems like it doesn’t support ingress rule with destination range, neither egress rule with source range.
 
https://github.com/terraform-google-modules/terraform-google-network/blob/v7.0.0/modules/firewall-rules/main.tf#L24
  source_ranges           = each.value.direction == "INGRESS" ? each.value.ranges : null
  destination_ranges      = each.value.direction == "EGRESS" ? each.value.ranges : null
***

imrannayer:
Fix #461


imrannayer:
@g-awmalik can you plz review?

modules/firewall-rules/variables.tf
```diff
@@ -25,28 +25,37 @@ variable "network_name" {
...
-    description             = optional(string)
-    direction               = optional(string)
-    priority                = optional(number)
-    ranges                  = optional(list(string))
+    description             = optional(string, null)
+    direction               = optional(string, "INGRESS")
+    priority                = optional(number, null)
+    ranges                  = optional(list(string), [])
+    destination_ranges      = optional(list(string), [])
+    source_ranges           = optional(list(string), [])
```
g-awmalik, line 37: It's confusing to see 3 range attributes although I understand that we're trying to not create a breaking change.

What do you think about introducing two new variables i.e. `ingress_rules` & `egress_rules` instead? By default we can redirect `rules` to the appropriate "rules" var i.e. `ingress_rules` or `egress_rules` based on the direction and use the other one for opposite direction. And add to `rules` description that it is deprecated.

This way, customers using `rules` will be fine for now while new customers could specifically use `ingress_rules` and/or `egress_rules` as needed. This will eliminate the need for custom validation for `rules` as well.

PR #464, opened 2023-06-24
feat: add destination_ranges and source_ranges in firewall rules

### LINKED ISSUE #461: Firewall rules for terraform-google-network module

imrannayer:
fix #461


modules/firewall-rules/main.tf
```diff
@@ -14,8 +14,14 @@
  * limitations under the License.
  */
 
+locals {
+  ingress_rules = { for rules in var.ingress_rules : rules.name => merge(rules, { direction = "INGRESS" }) }
```
g-awmalik, line 18: nit:

```suggestion
  ingress_rules = { for rule in var.ingress_rules : rule.name => merge(rule, { direction = "INGRESS" }) }
```

modules/firewall-rules/main.tf
```diff
@@ -14,8 +14,14 @@
...
+  egress_rules  = { for rules in var.egress_rules : rules.name => merge(rules, { direction = "EGRESS" }) }
```
g-awmalik, line 19: nit: as I was reading it i got confused for a second why all "rules" when it's being applied to each rule.

```suggestion
  egress_rules  = { for rule in var.egress_rules : rule.name => merge(rule, { direction = "EGRESS" }) }
```

examples/firewall-rules/main.tf
g-awmalik, line 1: Can we call this example "bidirectional-firewall-rules"? Something to indicate that we're creating both ingress and egress rules here.

examples/firewall-rules/versions.tf
g-awmalik, line 1: Don't think you need versions.tf in the example.

g-awmalik (approved):
Thanks for updating this. LGTM except for a few nits.

examples/firewall-rules/versions.tf
imrannayer: versions.tf is in every example of this and other modules. Are we planning to remove them from examples?

PR #468, opened 2023-06-30
refactor: updated the variables.tf to acomodate the change

Rashmi0986:
Updating this file to acomodate the network_name in the Main.tf file


imrannayer:
@Rashmi0986 we hard code values in our examples as the examples are also used for testing the code. You can make changes locally.

PR #473, opened 2023-08-01
feat: update module to remove deprecated rules field in firewall module

rayanebel:
* remove usage of `rules` variable in network module for the firewall part to use the newest variables: `ingress_rules` and `egress_rules`

This PR will contain a breaking changes and users will have to replace the `rules` field in the `firewall_rules` variable to remove `ranges` property and replace it with `source_ranges` or `destination_ranges`

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ Then perform the following commands on the root folder:
 
 | Name | Description |
 |------|-------------|
+| egress\_rules | n/a |
+| ingress\_rules | n/a |
 | network | The created network |
 | network\_id | The ID of the VPC being created |
 | network\_name | The name of the VPC being created |
--- a/main.tf
+++ b/main.tf
@@ -55,13 +55,14 @@ module "routes" {
 	Firewall rules
  *****************************************/
 locals {
-  rules = [
+  ingress_rules = [
     for f in var.firewall_rules : {
       name                    = f.name
       direction               = f.direction
       priority                = lookup(f, "priority", null)
       description             = lookup(f, "description", null)
-      ranges                  = lookup(f, "ranges", null)
+      source_ranges           = lookup(f, "source_ranges", null)
+      destination_ranges      = lookup(f, "destination_ranges", null)
       source_tags             = lookup(f, "source_tags", null)
       source_service_accounts = lookup(f, "source_service_accounts", null)
       target_tags             = lookup(f, "target_tags", null)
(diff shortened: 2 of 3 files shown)


imrannayer:
@rayanebel can you plz handle it same way as [firewall module is handling it](https://github.com/terraform-google-modules/terraform-google-network/blob/e17e051b9e7f4e9dcd967d1c63f539fd998b7d06/modules/firewall-rules/main.tf#L24). If user provides value for `rule`, module will use it otherwise it will use `ingress/egress` values. This way we can avoid breaking change and keep `rule` as deprecated variable.

rayanebel:
> @rayanebel can you plz handle it same way as [firewall module is handling it](https://github.com/terraform-google-modules/terraform-google-network/blob/e17e051b9e7f4e9dcd967d1c63f539fd998b7d06/modules/firewall-rules/main.tf#L24). If user provides value for `rule`, module will use it otherwise it will use `ingress/egress` values. This way we can avoid breaking change and keep `rule` as deprecated variable.

Hi @imrannayer 
Thanks for your reply. To be sure of what I have to do. Do you want from me to create two dedicated variables in the main module to have :

- `var.firewall_rules` which will point to `rules` in firewall module
- `var.firewall_ingress_rules` which will point to `ingress_rules` in firewall module
- `var.firewall_egress_rules` which will point to `egress_rules` in firewall module

imrannayer:
> > @rayanebel can you plz handle it same way as [firewall module is handling it](https://github.com/terraform-google-modules/terraform-google-network/blob/e17e051b9e7f4e9dcd967d1c63f539fd998b7d06/modules/firewall-rules/main.tf#L24). If user provides value for `rule`, module will use it otherwise it will use `ingress/egress` values. This way we can avoid breaking change and keep `rule` as deprecated variable.
> 
> Hi @imrannayer Thanks for your reply. To be sure of what I have to do. Do you want from me to create two dedicated variables in the main module to have :
> 
> * `var.firewall_rules` which will point to `rules` in firewall module (no changes for the users)
> * `var.firewall_ingress_rules` which will point to `ingress_rules` in firewall module
> * `var.firewall_egress_rules` which will point to `egress_rules` in firewall module

Can you plz check how firewall-rules module is handling it. If existing rule variable is pass behavior remains the same. Add two more variables and if length of existing rules variable is 0 then use the new variables.

PR #478, opened 2023-08-02
fix: network attachment error in firewall policy sub-module

### LINKED ISSUE #477: Duplicate name for attachment in network firewall policy sub-module
### TL;DR

When multiple VPC ID passed in target_vpc variable it throws an error:

```
googleapi: Error 400: Invalid value for field 'resource.name': 'regional-firewall-policy-mn70tg-us-central1'. An association with that name already exists
```

### Expected behavior

_No response_

### Observed behavior

_No response_

### Terraform Configuration

```hcl
module "firewal_policy" {
  source      = "../../modules/network-firewall-policy"
  project_id  = var.project_id
  policy_name = "${local.prefix}-firewall-policy-${random_string.random_suffix.result}"
  description = "test ${local.prefix} firewall policy"
  target_vpcs = [
    "projects/${var.project_id}/global/networks/${local.prefix}-network",
    "projects/${var.project_id}/global/networks/${local.prefix}-network-backup",
  ]
}
```
```


### Terraform Version

```sh
Terraform v1.5.1
on darwin_amd64
```

### Additional information

_No response_

imrannayer:
fix #477

### DIFF
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -168,7 +168,7 @@ steps:
   args: ['/bin/bash', '-c', 'cft test run TestPrivateServiceConnect --stage teardown --verbose']
 - id: converge global-firewall-policy
   waitFor:
-    - create all
+    - destroy private-service-connect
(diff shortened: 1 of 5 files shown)


bharathkkb (approved):
Since name is changing I suspect this will result in a delete + recreate but not marking as breaking since I don't think user action is needed

Issue #476, opened 2023-08-01
Role values no longer supported


montmanu:
### TL;DR

As of https://github.com/terraform-google-modules/terraform-google-network/pull/467, setting `role` on a `subnets` list item is no longer supported despite that usage being included in the [README's example](https://github.com/terraform-google-modules/terraform-google-network/blob/e17e051b9e7f4e9dcd967d1c63f539fd998b7d06/modules/subnets/README.md?plain=1#L34-L35).

### Expected behavior

No changes detected to the `role` field of the managed subnet resource after upgrading to `7.2`

### Observed behavior

State change detected for the`role` field of the managed subnet resource after upgrading to `7.2`.. `"ACTIVE" -> null`

```hcl
~ resource "google_compute_subnetwork" "subnetwork" {
    id                         = "projects/***/regions/***/subnetworks/***"
    name                       = "***"
    - role                     = "ACTIVE" -> null
    # (11 unchanged attributes hidden)
}
```

### Terraform Configuration

```hcl
module "vpc" {
  count = length(var.private_subnets) > 0 ? 1 : 0
  source  = "terraform-google-modules/network/google"
  version = "~> 7.1"
  project_id   = module.project.project_id
  network_name = var.name
  subnets = var.private_subnets # where one of the subnets items here has the `role` field set to `"ACTIVE"`
  secondary_ranges = var.secondary_ranges
}
```

### Terraform Version

```sh
$ terraform version
Terraform v1.5.2
on linux_amd64
+ provider registry.terraform.io/datadog/datadog v3.7.0
+ provider registry.terraform.io/hashicorp/google v4.73.1
+ provider registry.terraform.io/hashicorp/google-beta v4.73.1
+ provider registry.terraform.io/hashicorp/null v3.2.1
+ provider registry.terraform.io/hashicorp/random v3.5.1
+ provider registry.terraform.io/hashicorp/time v0.9.1
+ provider registry.terraform.io/hashicorp/vault v3.1.1
```

### Additional information

May also be an issue for the `purpose` field (or any other fields that were not included in the new type definition for the `subnets` variable).

parkedwards:
we are seeing this as well

saez0pub:
Hello,

I've also encountered this bug.
Waiting for a new release with a fix, I've locked the version to 7.1.0.
I've proposed the PR #479 , I hope it will be accepted 🤞.

imrannayer:
[...]

PR #479, opened 2023-08-02
fix(subnets): permachange on subnets

saez0pub:
missing properties in subnet type
Fixes https://github.com/terraform-google-modules/terraform-google-network/issues/476


imrannayer:
Fixed in #481

PR #515, opened 2023-11-02
fix!: bumping tf to v1.3 for fabric-net-firewall

g-awmalik:


imrannayer:
@g-awmalik should we consider TF version upgrade as breaking change?

g-awmalik:
> @g-awmalik should we consider TF version upgrade as breaking change?

Yes we should.

imrannayer (approved):

PR #530, opened 2023-12-14
fix: src_threat_intelligences is missing from the variable object definition.

joaquinza:
src_threat_intelligences is missing from the variable object definition.


imrannayer:
@joaquinza thx for the PR. CLA needs to be signed first.

Thanks

joaquinza:
> @joaquinza thx for the PR. CLA needs to be signed first.
> 
> Thanks

I tried signing it (and when I click on the links, it looks like signed) But for some reason it does not show like signed in the checks. Also I noticed that the check has not re-triggered since the original trigger.

joaquinza:
I added the readme modified by `make generate_docs` but the lint seems to be waiting on something.

imrannayer (approved):

PR #540, opened 2024-01-17
chore(examples): changed provider from google-beta to google

ajoisa:
Removed provider from main.tf and changed the provider name from google-beta to google in versions.tf


apeabody (approved):
Thanks for the contribution @ajoisa!

PR #541, opened 2024-01-17
chore(example): changed provider from google-beta to google

ajoisa:
Removed provider in main.tf and changed provider in versions.tf


apeabody:
merged https://github.com/terraform-google-modules/terraform-google-network/pull/540

PR #568, opened 2024-08-06
feat: Change serverless-connector input type and update metadata

q2w:
This PR makes below changes,

1. Updates input variable type for `vpc_connectors` from `list(map(string))` to `list(object({}))`
2. Updates metadata


modules/vpc-serverless-connector-beta/variables.tf
```diff
@@ -20,7 +20,18 @@ variable "project_id" {
...
 
 variable "vpc_connectors" {
-  type        = list(map(string))
+  type = list(object({
+    name              = string,
+    region            = string,
+    network           = optional(string, null),
+    subnet_name       = optional(string, null),
+    ip_cidr_range     = optional(string, null),
+    host_project_name = optional(string, null),
```
bharathkkb, line 29: I think this is the test failure reason
```suggestion
    host_project_id = optional(string, null),
```

q2w: Done!

modules/vpc/metadata.yaml
```diff
@@ -19,121 +19,146 @@ metadata:
...
-  - name: network_self_link
-    description: The URI of the VPC being created
-  - name: project_id
-    description: VPC project id
-  roles:
-  - level: Project
+  info:
+    title: Terraform VPC Module
+    source:
+      repo: https://github.com/q2w/terraform-google-network.git
```
bharathkkb, line 25: nit: here and a few places
```suggestion
      repo: https://github.com/terraform-google-modules/terraform-google-network.git
```

bharathkkb (approved):

modules/vpc/metadata.yaml
q2w: Thanks! Done.

PR #569, opened 2024-08-07
feat: Add new submodule for service networking

q2w:
This PR makes below changes,

1. Adds a new submodule for service-networking
2. Adds new example and test for the submodule


modules/service-networking/main.tf
```diff
@@ -0,0 +1,49 @@
...
+  purpose       = var.address_purpose
+  address_type  = var.address_type
+  prefix_length = var.address_prefix_length
+  network       = var.network_id
+}
+
+resource "google_service_networking_connection" "default" {
+  network                 = var.network_id
+  service                 = "servicenetworking.googleapis.com"
+  reserved_peering_ranges = [google_compute_global_address.global_address.name]
```
imrannayer, line 29: @q2w thx for the PR. It is possible to create one service networking with multiple global address ranges. If we run out of IP address we can add more by adding more IP address ranges in same service networking object. You cant create multiple instance of specific service networking which is why `reserved_peering_ranges` is a list of string. Can you plz define global address as an object with multiple global address and attach them all in `reserved_peering_ranges` list.
Also plz make `service` a variable as there are different possible values. Here is an [example](https://github.com/GoogleCloudPlatform/terraform-google-netapp-volumes/blob/main/examples/simple_example/main.tf)

q2w: Thanks for suggestion! Have included both the recommendations!

modules/service-networking/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  purpose       = each.value.purpose
+  address_type  = each.value.type
+  prefix_length = each.value.prefix_length
+  network       = var.network.id
+}
+
+resource "google_service_networking_connection" "default" {
+  network                 = var.network.id
+  service                 = var.service
+  reserved_peering_ranges = [for name, _ in google_compute_global_address.global_addresses : name]
```
imrannayer, line 30: You can write this as follows:
```
[for name in google_compute_global_address.global_addresses : name]
```

q2w: Done!

imrannayer: @q2w I apologize. Just realized it is an object. When u have one variable it will get values and u r looking for the key. 

you can rever this change.

q2w: Done.

modules/service-networking/metadata.yaml
```diff
@@ -0,0 +1,170 @@
...
+kind: BlueprintMetadata
+metadata:
+  name: terraform-google-network-service-networking
+  annotations:
+    config.kubernetes.io/local-config: "true"
+spec:
+  info:
+    title: Terraform Google service networking
+    source:
+      repo: https://github.com/q2w/terraform-google-network.git
```
bharathkkb, line 25:
```suggestion
      repo: https://github.com/terraform-google-modules/terraform-google-network.git
```

modules/service-networking/variables.tf
```diff
@@ -0,0 +1,85 @@
...
+    prefix_length : optional(number, 16)
+  }))
+}
+
+variable "network" {
+  description = "Network details including name and id"
+  type = object({
+    name = optional(string, null),
+    id   = string
+  })
```
bharathkkb, line 37: Why do we need both network name and id? Doesn't id of format `projects/{{project}}/global/networks/{{name}}` work in call cases?

modules/service-networking/main.tf
```diff
@@ -0,0 +1,50 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+resource "google_compute_global_address" "global_addresses" {
+  for_each      = { for address in var.global_addresses : address.name => address }
+  project       = var.project_id
+  name          = each.value.name
+  purpose       = each.value.purpose
```
bharathkkb, line 21: IIUC for this private service access usecase isn't it always `VPC_PEERING`?

modules/service-networking/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  address_type  = each.value.type
```
bharathkkb, line 22: Shouldn't we also be passing in the `address`?

modules/service-networking/README.md
```diff
@@ -0,0 +1,28 @@
+# Terraform Google service networking
+
+This module creates global network address and a service networking
```
bharathkkb, line 3: Could we add some more context on the uses cases? IIUC this module would be used for private service access. Is there any other usecase cc @imrannayer

modules/service-networking/main.tf
q2w: I have used VPC_PEERING as default value. Should i remove the variable and use VPC_PEERING directly?

modules/service-networking/main.tf
q2w: Added address. This is optional, GCP chooses a valid one if not provided. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_global_address#address

modules/service-networking/README.md
q2w: Updated the module README and example README

modules/service-networking/variables.tf
q2w: Made changes to use network name to derive network id

modules/service-networking/main.tf
imrannayer: @q2w In case of service networking it will always be vpc_peering.

imrannayer:
@q2w @bharathkkb how are we going to handle this sub-module and [current sub-module](https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/modules/private_service_access)?

They both are same. Although Current module has a flaw that it accepts one IP address instead of list of ip addresses.

q2w:
> @q2w @bharathkkb how are we going to handle this sub-module and [current sub-module](https://github.com/terraform-google-modules/terraform-google-sql-db/tree/master/modules/private_service_access)?
> 
> They both are same. Although Current module has a flaw that it accepts one IP address instead of list of ip addresses.


@imrannayer Thanks for pointing to the current module! Before creating this PR i tried to check if some module exist which can create service networking. I missed finding out private_service_access sub-module. IMHO the current submodule can fulfill the requirements. In future we can update the current module to take multiple addresses. I would defer to @bharathkkb for final call on the current PR.

bharathkkb:
Good catch @imrannayer, yes let's close this and update current module as Imran pointed out. Sorry for the churn @q2w!

PR #622, opened 2025-06-03
fix!: remove hardcoded gcp-beta in vpc sub-module

mkhtse:
The use of google-beta was added to accommodate network_profile at the time.  It's now GA and the hard-coded beta provider causes the return of URL for nework_profile  `https://www.googleapis.com/compute/beta/projects/.......` when the actual URL from the infrastructure is  
`networkProfile: https://www.googleapis.com/compute/v1/projects` ( v1 instead of beta)
you can validate that by running gcloud compute networks describe <network>

### DIFF
--- a/modules/vpc/main.tf
+++ b/modules/vpc/main.tf
@@ -18,7 +18,6 @@
 	VPC configuration
  *****************************************/
 resource "google_compute_network" "network" {
-  provider                                  = google-beta
   name                                      = var.network_name
   auto_create_subnetworks                   = var.auto_create_subnetworks
   routing_mode                              = var.routing_mode


mkhtse:
> @mkhtse thx for the PR. We decided to keep beta provider in case if we need to add any other beta feature in future..

Has any report the conflict related to the URL for network_profile returns beta instead of v1?  

This is the problem i'm facing. we work with google, they applied a network profile to an existing VPC network. To sync the config I had to remove the resource from state and re-import it.  The issue occurs when I match my TF config with the state file and infrastructure, where the URL return beta when it should be v1.   This causes  Force replacement for the network object and all the resources references the vpc network subsequently.

PR #657, opened 2025-12-26
feat(TPG>6.42)!: Add policy_type variable to network firewall policy

SwarnaBharathiMantena:
The `policy_type` variable is added to the network firewall policy. 

The `policy_type` variable is used to determine which resources (networks) the policy can be associated with. A policy can be associated with a network only if the network has the matching policyType in its network profile. Different policy types may support some of the Firewall Rules features. Possible values are: VPC_POLICY, RDMA_ROCE_POLICY.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [13.1.0](https://github.com/terraform-google-modules/terraform-google-network/compare/v13.0.1...v13.1.0) (2026-01-22)
+
+
+### Features
+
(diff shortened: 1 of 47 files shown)


SwarnaBharathiMantena:
Based on https://github.com/terraform-google-modules/terraform-google-network/blob/main/CONTRIBUTING.md#execution, I ran `make generate_docs`. This added changes to files `modules/network-connectivity-center/metadata.display.yaml` and `modules/private-service-connect-producer/metadata.display.yaml` which do not seem aligned to my intended change. 

I suspect it could be some missing pieces from earlier contributions.
I will revert changes on those files once I get a thumbs up on this.

imrannayer:
@SwarnaBharathiMantena can u plz resolve conflicts?

PR #693, opened 2026-04-30
feat: support reserved_internal_range in secondary_ranges

eoin-white-arista:
## Problem

The `reserved_internal_range` field was added to `google_compute_subnetwork.secondary_ip_range` in the Google provider v6.0.0. It is `Optional` but not `Computed`, meaning if it is present in GCP but absent from config, Terraform will plan to null it out.

GKE sets `reserved_internal_range` on VPC-native subnet secondary ranges when provisioning node pools — outside of Terraform. Because the `secondary_ranges` variable type only included `range_name` and `ip_cidr_range`, and the dynamic block only passed those two fields, Terraform would consistently plan to remove `reserved_internal_range` from any GKE-managed secondary range. GCP blocks this operation while nodes are attached to the subnet.

## Fix

- Added `reserved_internal_range = optional(string)` to the `secondary_ranges` variable type in `modules/subnets/variables.tf`, `modules/subnets-beta/variables.tf`, and the root `variables.tf`
- Passed `reserved_internal_range` through in the `secondary_ip_range` dynamic block in `modules/subnets/main.tf` and `modules/subnets-beta/main.tf`

Callers who manage GKE-provisioned secondary ranges can now set `reserved_internal_range` in their config to match the GCP-managed value, preventing destructive diffs.

## Testing

Validated against a real GKE cluster with active node pools.

**Before this fix**, `terraform plan` produced a destructive diff on subnets with GKE-managed secondary ranges:

```
  ~ resource "google_compute_subnetwork" "subnetwork" {
      ~ secondary_ip_range {
          ~ reserved_internal_range = "https://networkconnectivity.googleapis.com/v1/projects/my-project/locations/global/internalRanges/gke-my-cluster-services-xxxx" -> null
            range_name              = "gke-my-cluster-services"
            # (other attributes unchanged)
        }
    }
```

This apply would fail with GCP rejecting the removal of `reserved_internal_range` while nodes were attached.

**After this fix**, with the caller supplying the correct `reserved_internal_range` value in config, `terraform plan` produces no resource-level changes:

```
  ~ output "subnets_secondary_ranges" {
      ~ value = {
          ~ "europe-west1/my-subnet" = [
              ~ {
                  + reserved_internal_range = "https://networkconnectivity.googleapis.com/v1/projects/my-project/locations/global/internalRanges/gke-my-cluster-services-xxxx"
                    range_name              = "gke-my-cluster-services"
                    # (other attributes unchanged)
                }
            ]
        }
    }

Plan: 0 to add, 0 to change, 0 to destroy.
```

The only change is the output value in state being updated to reflect the field Terraform was previously ignoring — no real infrastructure changes.

## Notes

This overlaps with PR #640 which proposes the same fix. That PR is currently blocked on an integration test failure unrelated to the core change, and is based on an older version of the module. This PR applies the minimal fix cleanly against current `main`.

The `check_tflint` failure in CI is pre-existing on `main` (in `examples/network_connectivity_center/main.tf`) and unrelated to this change.


eoin-white-arista:
@imrannayer - if you could review please 🙏

nick-henry-arista (approved):
LGTM, thanks @eoin-white-arista

imrannayer:
@eoin-white-arista can you plz update versions.tf file with min version 6.0. 

https://github.com/terraform-google-modules/terraform-google-network/blob/ae68fe464b0682b4d4df91144089cb81ddd1bf75/modules/subnets-beta/versions.tf#L23

eoin-white-arista:
> @eoin-white-arista can you plz update versions.tf file with min version 6.0.
> 
> https://github.com/terraform-google-modules/terraform-google-network/blob/ae68fe464b0682b4d4df91144089cb81ddd1bf75/modules/subnets-beta/versions.tf#L23

Done @imrannayer  & thank you 🙇 . Pushed as d2eea91.

imrannayer (approved):

PR #712, opened 2026-08-26
feat: add support for other Google universes

daniel-cit:
This PR adds support for other Google universes in the configuration of the Private Service Connect module


modules/foundation/network/variables.tf
```diff
@@ -296,3 +296,22 @@ variable "bgp_inter_region_cost" {
   description = "Specifies the BGP inter-region cost mode. Valid values are `DEFAULT` or `ADD_COST_TO_MED`."
   default     = null
 }
+
+variable "universe_domain" {
```
apeabody, line 300: Should these have the same validation blocks as PSC module?

apeabody (approved):
Thanks @daniel-cit!

modules/foundation/network/variables.tf
daniel-cit: added the validation form PSC for clarity

daniel-cit: @apeabody Could you please marge this PR to go with the new foundation network module

